PR #334, opened 2023-06-15
Fix the labeled stats shown for labeled examples

rajasbansal:


src/autolabel/labeler.py
```diff
@@ -127,7 +127,7 @@ def run(
         for current_index in track_with_stats(
             indices,
             postfix_dict,
-            total=len(inputs),
+            total=len(inputs) - current_index,
```
Abhinav-Naikawadi, line 130: should this be "total=len(inputs) - current_index - 1"?

Abhinav-Naikawadi: if current_index is the number of elements that have been completed so far?

rajasbansal: Current index is the latest example which has *not* been labeled. So, if the first five examples have been labeled, the index would be 5 (0 indexing)

Abhinav-Naikawadi: ah okay never mind then!

Abhinav-Naikawadi (approved):

PR #339, opened 2023-06-16
add support for new OpenAI models, update pricing

nihit:
Support for gpt-3.5-turbo-0613, gpt-3.5-turbo-16k and updated pricing for gpt-3.5-turbo models

Tested civil comments with the following model config:
```
 "model": {
        "provider": "openai",
        "name": "gpt-3.5-turbo-0613"
    }
```

### DIFF
--- a/tests/unit/llm_test.py
+++ b/tests/unit/llm_test.py
@@ -77,7 +77,6 @@ def test_gpt35_label(mocker):
     )
     x = model.label(prompts)
     assert [i[0].text for i in x[0].generations] == ["Answers", "Answers"]
-    assert x[1] == approx(1.2e-05, rel=1e-3)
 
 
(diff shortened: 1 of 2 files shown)


tests/unit/llm_test.py
```diff
@@ -77,7 +77,6 @@ def test_gpt35_label(mocker):
     )
     x = model.label(prompts)
     assert [i[0].text for i in x[0].generations] == ["Answers", "Answers"]
-    assert x[1] == approx(1.2e-05, rel=1e-3)
```
nihit, line 80: this function checks for labeling only. we should test for pricing only it the following test after this one

nihit: cleaning up

DhruvaBansal00 (approved):

Issue #269, opened 2023-06-11
[Bug]: When using confidence, row in the table with threshold = `-inf` is confusing


rishabh-bhargava:
**Describe the bug**
When we are using Autolabel to compute confidence, the first row shows a threshold value = `-inf`. This is confusing to the user, since the next value is the max value. We already show the minimum threshold value at the bottom of the table, so we can skip the first row. 

**Expected behavior**
Remove the first row from the table, since it is not helpful. 

**Screenshots**
<img width="631" alt="image" src="https://github.com/refuel-ai/autolabel/assets/1568137/8df9b0aa-29ac-408d-bba9-785c7b764aeb">

Issue #328, opened 2023-06-15
[Bug]: When confidence is not specified, we produce a threshold = -inf, which is confusing


rishabh-bhargava:
**Describe the bug**
Even when `compute_confidence` is not set to `True`, we end up showing a threshold value of -inf, which is confusing to the user. We should hide that column. 

**To Reproduce**
Don't specify `compute_confidence` in the labeling config. 

**Expected behavior**
Instead we should get rid of the threshold column when confidence is not being computed. 

**Screenshots**
See screenshot from [the docs website](https://docs.refuel.ai/guide/overview/tutorial-classification/#experiment-1-try-simple-labeling-guidelines):
<img width="721" alt="image" src="https://github.com/refuel-ai/autolabel/assets/1568137/9d86340f-3889-4c22-a393-8d84d3d03328">

PR #343, opened 2023-06-19
remove the -inf row with compute confidence

Tyrest:
#269


rishabh-bhargava:
@Tyrest what does the final output look like?

Tyrest:
Here is with `compute_confidence=True`
Screenshot 2023-06-19 at 1 18 18 PM

Here is with `compute_confidence=False`
Screenshot 2023-06-19 at 1 19 02 PM

Tyrest:
This also covers #328

rishabh-bhargava (approved):

PR #345, opened 2023-06-19
Check if the LLM label is in the labels list

Tyrest:
If the LLM response does not correspond to a label in the labels list (only for `classification` and `entity_matching` tasks) then we will mark `successfully_labeled` as `False` for the row and print the warning as shown below.
Screenshot 2023-06-19 at 3 35 28 PM

### DIFF
--- a/src/autolabel/tasks/base.py
+++ b/src/autolabel/tasks/base.py
@@ -91,6 +91,7 @@ class BaseTask(ABC):
                 TaskType.CLASSIFICATION,
                 TaskType.ENTITY_MATCHING,
             ]:
+                logger.warning(f"LLM response is not in the labels list")
                 successfully_labeled = llm_label in self.config.labels_list()
             else:
                 successfully_labeled = True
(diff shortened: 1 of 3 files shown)


Tyrest:
This addresses issue #345

src/autolabel/tasks/base.py
```diff
@@ -86,8 +86,15 @@ def parse_llm_response(
             llm_label = self.NULL_LABEL_TOKEN
             logger.error(f"Error parsing LLM response: {response.text}")
         else:
-            successfully_labeled = True
             llm_label = completion_text.strip()
+            if self.config.task_type() in [
+                TaskType.CLASSIFICATION,
+                TaskType.ENTITY_MATCHING,
+            ]:
+                logger.warning(f"LLM response is not in the labels list")
```
Abhinav-Naikawadi, line 94: shouldn't we only be logging this warning when the label is not in the labels list (checked in next line)?

Abhinav-Naikawadi (changes requested):

src/autolabel/tasks/base.py
Tyrest: Oh yep, good catch! I'll change it now

rajasbansal (approved):

rajasbansal (commented):
could we add this for question answering as well where there are a set of answers for each question?

Abhinav-Naikawadi (approved):

Issue #316, opened 2023-06-15
[Feature Request]: Exponential backoff for PaLM


rajasbansal:
**Is your feature request related to a problem? Please describe.**
Currently we retry PaLM requests without any backoff i.e they are sent as soon as we error out in the other request.

**Describe the solution you'd like**
Exponential backoff similar to the refuel LLM

PR #347, opened 2023-06-20
added retry with tenacity for palm

Tyrest:
Since `langchain` does not support it, replaced the retry loop with a more robust solution from `tenacity` similar to how it is already implemented in `RefuelLLM`. #316


nihit (approved):
looks good! please test with one of the benchmark datasets/examples before merging

Tyrest:
Here's what it looks like when running Civil Comments with PaLM's `text-bison@001`'s quota capped at 20 / min.
Screenshot 2023-06-20 at 12 17 15 PM

rajasbansal (approved):

PR #354, opened 2023-06-21
added support for hugging face datasets

Tyrest:
We can now load datasets from Huggingface as show below
```python
dataset = load_dataset("lex_glue", "ledgar")

test_dataset = dataset["test"]
test_dataset = map_label_to_string(test_dataset, "label")

test_dataset = test_dataset.rename_column("text", "example")

ledgar_path = Path("../autolabel/examples/ledgar")
with open(ledgar_path / "config_ledgar.json", "r") as f:
    config = json.load(f)
if "few_shot_examples" in config["prompt"] and isinstance(config["prompt"]["few_shot_examples"], str):
    config["prompt"]["few_shot_examples"] = str(ledgar_path / config["prompt"]["few_shot_examples"])
agent = LabelingAgent(config)

agent.plan(test_dataset, max_items=8)
agent.run(test_dataset, max_items=8)
```

### DIFF
--- a/src/autolabel/dataset_loader.py
+++ b/src/autolabel/dataset_loader.py
@@ -218,11 +218,11 @@ class DatasetLoader:
             max_items (int, optional): max number of items to read. Defaults to None.
             start_index (int, optional): start index to read from. Defaults to 0.
         """
-        self.dat = pd.DataFrame(
-            dataset[
-                start_index : max_items if max_items and max_items > 0 else len(dataset)
-            ]
-        )
+        dataset.set_format("pandas")
+        self.dat = dataset[
+            start_index : max_items if max_items and max_items > 0 else len(dataset)
+        ]
+
         self.inputs = self.dat.to_dict(orient="records")
         self.gt_labels = (
             None
--- a/src/autolabel/dataset_loader.py
+++ b/src/autolabel/dataset_loader.py
@@ -3,6 +3,7 @@ from typing import Dict, List, Tuple, Union
 import logging
 import pandas as pd
 from sqlalchemy.sql.selectable import Selectable
+from datasets import Dataset
 
 
 from autolabel.configs import AutolabelConfig
@@ -36,6 +37,8 @@ class DatasetLoader:
 
         if isinstance(dataset, str):
             self._read_file(dataset, config, max_items, start_index)
+        elif isinstance(dataset, Dataset):
+            self._read_hf_dataset(dataset, config, max_items, start_index)
         elif isinstance(dataset, pd.DataFrame):
             self._read_dataframe(dataset, config, start_index, max_items)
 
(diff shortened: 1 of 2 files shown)


src/autolabel/dataset_loader.py
```diff
@@ -199,3 +202,32 @@ def _read_file(
...
+
+        Args:
+            dataset (Dataset): dataset object to read from
+            config (AutolabelConfig): config object
+            max_items (int, optional): max number of items to read. Defaults to None.
+            start_index (int, optional): start index to read from. Defaults to 0.
+        """
+        self.dat = pd.DataFrame(
+            dataset[
+                start_index : max_items if max_items and max_items > 0 else len(dataset)
```
nihit, line 223: do we need to do any sanity checks on the dataset object before attempting to convert to dataframe?

src/autolabel/labeler.py
```diff
@@ -66,17 +66,21 @@ def run(
             output_name: custom name of output CSV file
             start_index: skips annotating [0, start_index)
         """
+        dataset_loader = DatasetLoader(dataset, self.config, max_items, start_index)
 
         self.db.initialize()
         self.dataset = self.db.initialize_dataset(
-            dataset, self.config, start_index, max_items
+            dataset_loader.dat, self.config, start_index, max_items
```
nihit, line 73: can we access the dataframe with a helper method in dataset loader (vs directly accessing it here) ? for readability

src/autolabel/labeler.py
```diff
@@ -66,17 +66,21 @@ def run(
...
-
-        dataset_loader = DatasetLoader(dataset, self.config, max_items, start_index)
+        if isinstance(dataset, str):
+            csv_file_name = (
+                output_name
+                if output_name
+                else f"{dataset.replace('.csv','')}_labeled.csv"
+            )
+        else:
+            csv_file_name = "labeled.csv"
```
nihit, line 83: let's change this to contains a task name prefix_labeled.csv so that the output files from different tasks don't overwrite each other

src/autolabel/dataset_loader.py
Tyrest: [This](https://discuss.huggingface.co/t/converting-an-hf-dataset-to-pandas/12043) seems to be a more robust solution so I'll implement it here as well

src/autolabel/labeler.py
Tyrest: Can you give an example of what you want it to look like instead?

nihit: i was thinking just having a `def as_datafame()` method in the data loader that returns the internal pandas dataframe object, instead of the labeling agent accessing it directly

nihit (approved):

PR #366, opened 2023-06-22
Adding default cpu support. Removing max tokens for huggingface warnings

DhruvaBansal00:


src/autolabel/models/hf_pipeline.py
```diff
@@ -38,7 +38,9 @@ def __init__(self, config: AutolabelConfig, cache: BaseCache = None) -> None:
         # initialize HF pipeline
         tokenizer = AutoTokenizer.from_pretrained(self.model_name)
         quantize_bits = self.model_params["quantize"]
-        if quantize_bits == 8:
+        if not torch.cuda.is_available():
+            model = AutoModelForSeq2SeqLM.from_pretrained(self.model_name)
```
nihit, line 42: is model quantization only respected for GPU inference? makes sense but just want to confirm

nihit (approved):
lgtm otherwise

src/autolabel/models/hf_pipeline.py
DhruvaBansal00: Yeah https://github.com/TimDettmers/bitsandbytes/issues/40, quantization isn't really supported on cpu.

Issue #371, opened 2023-06-23
[Bug]: Malformed JSON config for starter task


yadavsahil197:
The JSON config for the starter task MovieSentimentReview is malformed.

**To Reproduce**
Saving the given json to a file and using it with `LabelingAgent("config.json")` complains because of python dict vs json handling of commas. Removing the additional comma so the config can be saved as a json file as well.
Also, there are typos and invalid newline character in the config.

Current Error: https://gist.github.com/yadavsahil197/3fd60c489a91b81b91d77579cd5d24ae

**Expected behavior**
The config should not have typos and be loaded as json file.

PR #372, opened 2023-06-23
Fix JSON config for starter task

yadavsahil197:
Fixes https://github.com/refuel-ai/autolabel/issues/371
Tested by loading the config into json file and running
`agent = LabelingAgent("config.json")`
and also using the dictionary directly
`agent = LabelingAgent(config)`


nihit (approved):
thanks for fixing this!

PR #375, opened 2023-06-23
Checks should run on PR commits and after push to main branch

yadavsahil197:
Previously we were running tests on PR but on the base commit on the target branch. Modified it to run the tests on the commit on the PR merge branch.
Also, the tests should only run when they are targeting main branch.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,32 +1,34 @@
 name: Tests
 
-on: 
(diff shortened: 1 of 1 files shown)


nihit (approved):

yadavsahil197:
The checks are failing because of the langchain, It'll be fixed in this PR https://github.com/refuel-ai/autolabel/pull/374

PR #361, opened 2023-06-22
added custom progress bar for get data

chiranthans23:
Maintaining the same progress bar as before but specifying the #bytes downloaded
image

PR #383, opened 2023-06-24
explicitly states it's bytes being downloaded in get_data.py

JTarakRam:
Here we can explicitly states it's bytes being downloaded


rishabh-bhargava:
@JTarakRam looks like this was already fixed and merged here: https://github.com/refuel-ai/autolabel/pull/361

going to close this PR, but can reopen if needed.

Issue #352, opened 2023-06-21
[Feature Request]: Support logit_bias to constrain LLM outputs 


nihit:
**Is your feature request related to a problem? Please describe.**
LLM might sometimes have trouble following output format instructions

**Describe the solution you'd like**
logit_bias: https://twitter.com/AAAzzam/status/1669753725093654565

nihit:
We can start with support for classification and entity matching tasks

nihit:
thanks @neildhruva for the suggestion!

DhruvaBansal00:
logit_bias is an OpenAI only  feature at the moment.
For huggingface equivalent, we can use this: https://huggingface.co/docs/transformers/main/en/internal/generation_utils#transformers.SequenceBiasLogitsProcessor

PR #385, opened 2023-06-26
Implemented logit bias

Tyrest:
This PR addresses #352. Not completely sure if this is the right place to implement it.


DhruvaBansal00:
Is implementing this for huggingface within the scope of this PR? Or is this only for supporting biased generation with OpenAI?

Tyrest:
Currently this is implemented only how OpenAI takes it (as a model parameter)

DhruvaBansal00 (approved):
LGTM!

rishabh-bhargava:
Need to make sure that adding `logit_bias` in tasks other than classification / entity matching throw an error.

Issue #413, opened 2023-06-30
[Bug]: .autolabel.db is written to ~


turian:
**Describe the bug**

The docs say:
"The database is saved in the same directory from where you run the LabelingAgent notebook and is named .autolabel.db."

However, the [code](https://github.com/refuel-ai/autolabel/blob/main/src/autolabel/database/engine.py) saves to the user's home directory.

**Expected behavior**
The code should save the db to the same directory from where you run the LabelingAgent notebook.

**Additional context**
Having one global SQLite database in ~ is a poor idea, since SQLite cannot handle multiple simultaneous writes (if you have several different labeling jobs going on).

PR #414, opened 2023-06-30
Fix engine path

turian:
Fixes https://github.com/refuel-ai/autolabel/issues/413

### DIFF
--- a/src/autolabel/database/engine.py
+++ b/src/autolabel/database/engine.py
@@ -4,7 +4,14 @@ from sqlalchemy.engine.base import Engine
 from os.path import join, expanduser
 
 DB_ENGINE = None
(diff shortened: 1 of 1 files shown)


rishabh-bhargava:
Thanks for opening the issue and this PR @turian. Curious: were you seeing a bottleneck in the number of successful writes to SQLite while labeling?

rishabh-bhargava (approved):

turian:
@rishabh-bhargava I didn't see bottlenecks, but I have been playing with sqlite for a while.

I was seeing that when I extended the max number of examples in my run from 1K to 5K that the first 1K took just as much time to run, as if the cache weren't getting hit.

PR #419, opened 2023-06-30
added cohere support

Tyrest:
Added [Cohere](https://cohere.com/) support that currently only supports the generation models, defaulting to `command`. To use it, set `config["model"]["provider"]` to `cohere`.

### DIFF
--- a/src/autolabel/models/cohere.py
+++ b/src/autolabel/models/cohere.py
@@ -1,0 +1,58 @@
+from functools import cached_property
+from typing import List, Optional
+import os
+
+import cohere
+from langchain.llms import Cohere
+from langchain.schema import LLMResult
+from langchain import PromptTemplate, LLMChain
+
+from autolabel.models import BaseModel
+from autolabel.configs import AutolabelConfig
+from autolabel.cache import BaseCache
+
+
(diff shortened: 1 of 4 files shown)


src/autolabel/models/cohere.py
```diff
@@ -0,0 +1,58 @@
...
+from autolabel.models import BaseModel
+from autolabel.configs import AutolabelConfig
+from autolabel.cache import BaseCache
+
+
+class CohereLLM(BaseModel):
+    # Default parameters for OpenAILLM
+    DEFAULT_MODEL = "command"
+    DEFAULT_MODEL_PARAMS = {
+        "max_tokens": 1000,
```
DhruvaBansal00, line 19: Is this max input tokens, max output tokens, or the sum of both? Let’s make sure we at least allow 2048 for input tokens and 512 for output tokens by default.

DhruvaBansal00 (approved):
LGTM

src/autolabel/models/cohere.py
Tyrest: This is the max output tokens, I'll make the change to 512 now!

PR #423, opened 2023-07-01
Fix open api key error handling

### LINKED ISSUE #421: [Bug]: Confusing error message when OpenAI API Key is missing
**Describe the bug**
When the OpenAI API key is missing, we see the following error message:

```
File ~/.pyenv/versions/3.8.13/envs/refuel-main/lib/python3.8/site-packages/langchain/load/serializable.py:64, in Serializable.__init__(self, **kwargs)
     63 def __init__(self, **kwargs: Any) -> None:
---> 64     super().__init__(**kwargs)
     65     self._lc_kwargs = kwargs

File ~/.pyenv/versions/3.8.13/envs/refuel-main/lib/python3.8/site-packages/pydantic/main.py:341, in pydantic.main.BaseModel.__init__()

ValidationError: 1 validation error for ChatOpenAI
__root__
  Did not find openai_api_key, please add an environment variable `OPENAI_API_KEY` which contains it, or pass  `openai_api_key` as a named parameter. (type=value_error)
```

This is an error from `langchain`, but `autolabel` doesn't allow passing `openai_api_key` as a named parameter anywhere. 

**To Reproduce**
Call `agent.plan` for a config that uses an OpenAI model without providing the API key as an env variable. 

**Expected behavior**
We should throw an error that simply says that the user should add the API key as an environment variable and not that it should be a named parameter.

chirag-manwani:
Fixes #421 

This PR adds a check for the presence of the OPENAI_API_KEY environment variable. Also it fixes a typo in the config.json in the README (and a couple of linting changes).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,5 @@
 <img alt="Refuel logo" src="https://raw.githubusercontent.com/refuel-ai/autolabel/main/docs/assets/Autolabel_blk_w_background.png">
 
-
 <h4 align="center">
   <a href="https://discord.gg/fweVnRx6CU">Discord</a> |
   <a href="https://twitter.com/RefuelAI">Twitter</a> |
@@ -54,7 +53,7 @@ Let's imagine we are building an ML model to analyze sentiment analysis of movie
         "labels": [
             "positive",
             "negative",
-            "neutral",
+            "neutral"
         ],
         "few_shot_examples": [
             {
@@ -166,7 +165,6 @@ Check out our [technical report](https://www.refuel.ai/blog-posts/llm-labeling-t
 Check out our [public roadmap](https://github.com/orgs/refuel-ai/projects/15) to learn more about ongoing and planned improvements to the Autolabel library.
 
 We are always looking for suggestions and contributions from the community. Join the discussion on [Discord](https://discord.gg/fweVnRx6CU) or open a [Github issue](https://github.com/refuel-ai/autolabel/issues) to report bugs and request features.
-
 
 ## 🙌 Contributing
 
--- a/src/autolabel/models/openai.py
+++ b/src/autolabel/models/openai.py
@@ -11,6 +11,7 @@ from autolabel.models import BaseModel
 from autolabel.configs import AutolabelConfig
 from autolabel.cache import BaseCache
 
+import os
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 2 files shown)


chirag-manwani:
I considered wrapping the code between [line-91](https://github.com/refuel-ai/autolabel/blob/e73342bebb2936d2aeb1200d6b07da24f2e3dbbc/src/autolabel/models/openai.py#L91)  and [line-99](https://github.com/refuel-ai/autolabel/blob/e73342bebb2936d2aeb1200d6b07da24f2e3dbbc/src/autolabel/models/openai.py#L99) inside a try/except block and checking for the Validation error message for a missing API key(I am not completely sure if there is a better way), but in that case I thought the code should handle all sorts of Validation errors and not just this one. However since there can be many such Validation Errors, [where one Validation Error object may contain more than one error](https://github.com/pydantic/pydantic/blob/15d5e0b0abdd79fd63d34954e8dacaa9f1a255f8/pydantic/v1/error_wrappers.py#L58C31-L58C40), it seemed like a better idea to check for the API key separately. 

This was the code I thought would work but not sure how to handle other validation errors. 
```python
try:
    if self._engine == "chat":
        self.model_params = {**self.DEFAULT_PARAMS_CHAT_ENGINE, **model_params}
        self.llm = ChatOpenAI(model_name=self.model_name, **self.model_params)
    else:
        self.model_params = {
            **self.DEFAULT_PARAMS_COMPLETION_ENGINE,
            **model_params,
        }
        self.llm = OpenAI(model_name=self.model_name, **self.model_params)
except ValidationError as err:
    if err.__str__().find("Did not find openai_api_key") != -1:
        raise ValueError("OPENAI_API_KEY environment variable not set") from None
    else:
        raise err from None
```

chirag-manwani:
This is what it looks like. 

```python
from autolabel import LabelingAgent
agent = LabelingAgent(config='downloads/config.json')
```

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/packages/venv/env-1/lib/python3.11/site-packages/autolabel/labeler.py", line 49, in __init__
    self.llm: BaseModel = ModelFactory.from_config(self.config, cache=self.cache)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/packages/venv/env-1/lib/python3.11/site-packages/autolabel/models/__init__.py", line 58, in from_config
    return model_cls(config, cache)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/packages/venv/env-1/lib/python3.11/site-packages/autolabel/models/openai.py", line 103, in __init__
    raise ValueError("OPENAI_API_KEY environment variable not set") from None
ValueError: OPENAI_API_KEY environment variable not set
```

nihit (approved):
lgtm! thanks @chirag-manwani for this fix

nihit:
If you haven't already, also consider joining our Discord: https://discord.gg/fweVnRx6CU

PR #436, opened 2023-07-07
Support example selection using label diversity by random and by semantic similarity

Abhinav-Naikawadi:
Currently, we order the examples grouped by label, but might want to randomize the order of the selected examples.

### DIFF
--- a/src/autolabel/few_shot/label_diversity_example_selector.py
+++ b/src/autolabel/few_shot/label_diversity_example_selector.py
@@ -54,6 +54,7 @@ class LabelDiversityRandomExampleSelector(BaseExampleSelector, BaseModel):
         cls,
         examples: List[dict],
         label_key: str,
+        num_labels: int,
         k: int = 4,
     ) -> LabelDiversityRandomExampleSelector:
         """Create label diversity example selector using example list and embeddings.
@@ -66,7 +67,7 @@ class LabelDiversityRandomExampleSelector(BaseExampleSelector, BaseModel):
         Returns:
             The ExampleSelector instantiated
         """
-        return cls(k=k, examples=examples, label_key=label_key)
+        return cls(k=k, examples=examples, label_key=label_key, num_labels=num_labels)
 
 
 class LabelDiversitySimilarityExampleSelector(BaseExampleSelector, BaseModel):
@@ -123,6 +124,7 @@ class LabelDiversitySimilarityExampleSelector(BaseExampleSelector, BaseModel):
         label_key: str,
         embeddings: Embeddings,
         vectorstore_cls: Type[VectorStore],
+        num_labels: int,
         k: int = 4,
         input_keys: Optional[List[str]] = None,
         **vectorstore_cls_kwargs: Any,
@@ -157,4 +159,5 @@ class LabelDiversitySimilarityExampleSelector(BaseExampleSelector, BaseModel):
             k=k,
             input_keys=input_keys,
             label_key=label_key,
+            num_labels=num_labels,
         )
--- a/src/autolabel/few_shot/label_diversity_example_selector.py
+++ b/src/autolabel/few_shot/label_diversity_example_selector.py
@@ -1,6 +1,7 @@
 """Example selector that selects examples based on SemanticSimilarity."""
 from __future__ import annotations
 
+import math
 from itertools import groupby
 from operator import itemgetter
 from typing import Any, Dict, List, Optional, Type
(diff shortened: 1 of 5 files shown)


rishabh-bhargava:
What's a good test to evaluate how well these strategies work?

src/autolabel/few_shot/label_diversity_example_selector.py
```diff
@@ -0,0 +1,153 @@
...
+from pydantic import BaseModel, Extra
+
+
+def sorted_values(values: Dict[str, str]) -> List[Any]:
+    """Return a list of values in dict sorted by key."""
+    return [values[val] for val in sorted(values)]
+
+
+class LabelDiversityRandomExampleSelector(BaseExampleSelector, BaseModel):
+    """Example selector that selects examples based on SemanticSimilarity."""
```
nihit, line 20: update this

src/autolabel/few_shot/label_diversity_example_selector.py
```diff
@@ -0,0 +1,153 @@
...
+
+        extra = Extra.forbid
+        arbitrary_types_allowed = True
+
+    def add_example(self, example: Dict[str, str]) -> None:
+        self.examples.append(example)
+
+    def select_examples(self, input_variables: Dict[str, str]) -> List[dict]:
+        selected_examples = []
+        sorted_examples = sorted(self.examples, key=itemgetter(self.label_key))
```
nihit, line 40: i didn't quite get this step - if we're grouping by the label next, do we need to do this sorting?

src/autolabel/few_shot/label_diversity_example_selector.py
```diff
@@ -0,0 +1,153 @@
...
+    """Return a list of values in dict sorted by key."""
+    return [values[val] for val in sorted(values)]
+
+
+class LabelDiversityRandomExampleSelector(BaseExampleSelector, BaseModel):
+    """Example selector that selects examples based on SemanticSimilarity."""
+
+    examples: List[dict]
+    """A list of the examples that the prompt template expects."""
+    k: int = 1
```
nihit, line 24: "k" parameter in https://github.com/refuel-ai/autolabel/pull/436/files#diff-374f83319a2bfe5304bc20448393efad6a01c24c64d1cc276180d67141c99d65L61 is for the total number of few-shot examples. we need to map this to num examples per label by dividing it by number of labels?

src/autolabel/few_shot/label_diversity_example_selector.py
```diff
@@ -0,0 +1,153 @@
...
+        cls,
+        examples: List[dict],
+        label_key: str,
+        k: int = 4,
+    ) -> LabelDiversityRandomExampleSelector:
+        """Create label diversity example selector using example list and embeddings.
+
+        Args:
+            examples: List of examples to use in the prompt.
+            k: Number of examples to select per label
```
nihit, line 59: "k" in the upstream caller class is the total number of few-shot examples

src/autolabel/few_shot/label_diversity_example_selector.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+        return cls(k=k, examples=examples, label_key=label_key)
+
+
+class LabelDiversitySimilarityExampleSelector(BaseExampleSelector, BaseModel):
+    """ExampleSelector that selects examples based on label diversity, while choosing the most similar examples for each label"""
+
+    vectorstore: VectorStore
+    """VectorStore than contains information about examples."""
+    k: int = 4
```
nihit, line 73: why are we setting the num examples per label default value different here (k = 4 vs k = 1) in `LabelDiversityRandomExampleSelector` ?

nihit (approved):
1. please add this to few-shot prompting docs (can do as a followup PR)
2. would be good to understand how well does this work - let's take a couple of datasets from the few-shot experiment in https://www.refuel.ai/blog-posts/llm-labeling-technical-report to understand how well this does compared to a fixed set, and semantically similar set - I suggest at least sanity checking this with one dataset before merging.

lgtm otherwise

src/autolabel/few_shot/label_diversity_example_selector.py
Abhinav-Naikawadi: groupby just chunks consecutive groups with the same value so it requires the list to be sorted by the value for our use case (otherwise we'd have several chunks of each label interspersed)

nihit (approved):

PR #438, opened 2023-07-08
label diversity example selection docs

Abhinav-Naikawadi:


docs/guide/accuracy/few-shot.md
```diff
@@ -181,6 +181,64 @@ labels, df, metrics_list = agent.run('../examples/banking/test.csv')
...
+    }
+}
+```
+
+```py
+agent = LabelingAgent(config=config_label_diversity_similarity)
+labels, df, metrics_list = agent.run('../examples/civil_comments/test.csv', max_items=200)
+```
+
+For this run on the civil comments dataset, label diversity at random achieved 80% accuracy and label diversity with semantic similarity achieved 78% accuracy. For the same subset of data, the use of regular semantic similarity example selection obtained 72% accuracy, making for a significant improvement by using label diversity. Label diversity example selection strategies are likely best suited for labeling tasks with a small number of unique labels, which is the case for the civil comments dataset with only 2 labels. This is because equal representation of all the possible labels may be less likely to bias the LLM towards a particular label.
```
turian, line 240: Cool!

nihit (approved):

PR #441, opened 2023-07-10
Dataset validation fix for explanation generation

Abhinav-Naikawadi:
2 fixes here: 

1) read_file no longer exists in the DatasetLoader class so updated that

2) There was a subsequent set of errors due to dataset validation expecting an explanation column in the dataset (in seed.csv before explanation column was generated and in test.csv since explanation column isn't supposed to exist in the test dataset). Added a fix to not include the explanation column in all dataset validation. We do already have a separate check for the existence of the explanation column prior to example selection when appropriate here: https://github.com/refuel-ai/autolabel/blob/main/src/autolabel/labeler.py#L107

### DIFF
--- a/src/autolabel/data_loaders/__init__.py
+++ b/src/autolabel/data_loaders/__init__.py
@@ -1,18 +1,18 @@
 import logging
-import pandas as pd
-from tabulate import tabulate
 from typing import Dict, Union
-from datasets import Dataset
-from autolabel.data_loaders.read_datasets import (
+
+import pandas as pd
+from autolabel.configs import AutolabelConfig
+from autolabel.data_loaders.read_datasets import (  # SqlDatasetReader,
     AutolabelDataset,
     CSVReader,
-    JsonlReader,
-    HuggingFaceDatasetReader,
-    # SqlDatasetReader,
     DataframeReader,
+    HuggingFaceDatasetReader,
+    JsonlReader,
 )
 from autolabel.data_loaders.validation import TaskDataValidation
-from autolabel.configs import AutolabelConfig
+from datasets import Dataset
+from tabulate import tabulate
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


src/autolabel/data_loaders/validation.py
```diff
@@ -237,9 +241,11 @@ def validate(self, data: List[dict]) -> List[Dict]:
     def validate_dataset_columns(self, dataset_columns: List):
         """Validate columns
 
-        Valiate if the columns mentioned in example_template dataset are correct
-        and are contined within the columns of the dataset(seed.csv)
+        Validate if the columns mentioned in example_template dataset are correct
+        and are contained within the columns of the dataset(seed.csv)
         """
+        if self.explanation_column in dataset_columns:
+            dataset_columns.remove(self.explanation_column)
```
rajasbansal, line 248: this will modify the list being passed into the function - might be difficult for the caller to understand

Is it possible to remove this code from validate_dataset_columns and let expected_columns handle the removal of explanation column?

src/autolabel/data_loaders/__init__.py
```diff
@@ -1,18 +1,18 @@
 import logging
-import pandas as pd
-from tabulate import tabulate
 from typing import Dict, Union
-from datasets import Dataset
-from autolabel.data_loaders.read_datasets import (
+
+import pandas as pd
+from autolabel.configs import AutolabelConfig
+from autolabel.data_loaders.read_datasets import (  # SqlDatasetReader,
```
rajasbansal, line 6: do we need this?

rajasbansal (approved):

src/autolabel/data_loaders/__init__.py
Abhinav-Naikawadi: I think this was just a black formatting reordering of imports

rajasbansal (approved):

Issue #433, opened 2023-07-06
[Bug]: Retry remaining batch of prompts on failure


yadavsahil197:
**Describe the bug**
The labeling agent sends CHUNK_SIZE no of prompts for labeling to the LLM at a given time. If the LLM fails to label some of the prompts, the entire chunk of prompts are left unlabeled.

**To Reproduce**
A list of steps to reproduce the behavior:
Use an LLM with smaller context window and create a prompt example with large enough tokens, all prompts in the batch fail even if some of them have tokens within the context window of the model.

**Expected behavior**
The labeling agent should fail only the actual failure prompt and retry for the remaining prompts in the batch.

Abhinav-Naikawadi:
@Tyrest we can close this right? Done here: #443

PR #443, opened 2023-07-11
added individual retrying for failed batches

Tyrest:
If the batch fails, simply splits the batch into the individual prompts and then retries them.


Tyrest:
This is currently done synchronously.

Tyrest:
#433

yadavsahil197:
Does it make sense to move the retry logic into a wrapper method within the base model class?

yadavsahil197 (approved):
LGTM

PR #444, opened 2023-07-11
add support for older pinned gpt-3.5-turbo-0301

nihit:


rishabh-bhargava:
when we update docs, we will also want to add these older models + some of the 16k models to this providers list: https://docs.refuel.ai/guide/llms/llms/#provider-list

rishabh-bhargava:
`gpt-4-0314` and `gpt-4-32k-0314` as well?

rishabh-bhargava (approved):

PR #445, opened 2023-07-11
add multilabel classification to validation

Tyrest:


src/autolabel/data_loaders/validation.py
```diff
@@ -109,11 +109,41 @@ def validate(self, value: str):
...
+    def validate(self, value: str):
+        if value.startswith("[") and value.endswith("]"):
+            try:
+                seed_labels = eval(value)
+                if not isinstance(seed_labels, list):
+                    raise
+                unmatched_label = set(seed_labels) - self.labels_set
+                if len(unmatched_label) != 0:
+                    raise ValueError(
+                        f"labels: '{unmatched_label}' not in promt/labels provided in config "
```
rajasbansal, line 132: nit typo

src/autolabel/data_loaders/validation.py
```diff
@@ -109,11 +109,41 @@ def validate(self, value: str):
...
+
+    label_column: str
+    labels_set: set  # A Multilabel Classification Task should have a unique set of labels in config
+
+    def validate(self, value: str):
+        if value.startswith("[") and value.endswith("]"):
+            try:
+                seed_labels = eval(value)
+                if not isinstance(seed_labels, list):
+                    raise
```
rajasbansal, line 128: trying to understand how the empty raise works - from what i read it means that it raises the last exception that was handled. Given we are not in an except clause here, what would an empty raise mean here?
I would suggest raising a ValueError just for readability

Tyrest: I was just copying the other validation functions but it makes more sense to raise a `ValueError` here so I'll add that now.

rajasbansal (approved):
lgtm

Issue #409, opened 2023-06-29
[Bug]: Consolidate calculate f1


Tyrest:
**Describe the bug**
Currently, we compute f1 in two different places: `tasks/utils.py` and in `multilabel_classification.py` with `sklearn`. We should consolidate these functions into a single place and use that instead.

**Additional context**
See conversation about this [here](https://github.com/refuel-ai/autolabel/pull/405#pullrequestreview-1504612482)

nihit:
@rajasbansal to review

PR #453, opened 2023-07-12
consolidate compute f1 to single function in utils

Tyrest:
`compute_f1` now takes in the entire list of gt labels and llm labels instead of just one at a time. The binarization is also done in `compute_f1` for multilabel classification. This directly addresses #409.


Tyrest:
Now, the multilabel classification task evaluation table will look something like this:
Screenshot 2023-07-13 at 1 27 43 PM

rajasbansal (approved):
lgtm! Lets add f1 score as a metric for multi class classification as well similar to this. Can track this in a separate ticket if needed.

PR #454, opened 2023-07-13
Generate Multiple Predictions

DhruvaBansal00:
This PR adds the ability to generate and save multiple predictions from the following providers:

openai
huggingface
refuel


rishabh-bhargava:
@DhruvaBansal00 - what's the plan here?

Issue #396, opened 2023-06-27
Cache embedding calls in the  Autolabel cache


nihit:

nihit:
@Sardhendu

PR #457, opened 2023-07-13
cache embeddings in vector store

Tyrest:
Now autolabel caches the embeddings when `cache=True` in the same database where task runs are cached. This resolves the need to recompute all the embeddings every time.

This directly addresses #396

### DIFF
--- a/src/autolabel/few_shot/vector_store.py
+++ b/src/autolabel/few_shot/vector_store.py
@@ -4,6 +4,8 @@ import heapq
 from itertools import groupby
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Type
 
+from autolabel.database import create_db_engine
+
 import numpy as np
 import torch
 from langchain.docstore.document import Document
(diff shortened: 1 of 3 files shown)


src/autolabel/few_shot/vector_store.py
```diff
@@ -136,11 +141,78 @@ def __init__(
...
+            self._db_engine = None
+
+    def _get_embeddings(self, texts: Iterable[str]) -> List[List[float]]:
+        """Get embeddings from the database. If not found, compute them and add them to the database.
+        Args:
+            texts (Iterable[str]): Iterable of texts to embed.
+        Returns:
+            List[List[float]]: List of embeddings.
+        """
+        if self._db_engine:
```
nihit, line 166: if db_engine is not set (cache = False), what is the behavior of this function?

src/autolabel/few_shot/vector_store.py
```diff
@@ -136,11 +141,78 @@ def __init__(
...
+
+    def _add_embeddings(
+        self, texts: Iterable[str], embeddings: List[List[float]]
+    ) -> None:
+        """Save embeddings to the database.
+        Args:
+            texts (Iterable[str]): Iterable of texts.
+            embeddings (List[List[float]]): List of embeddings.
+        """
+        if self._db_engine:
```
nihit, line 204: similar question

nihit (approved):

Issue #460, opened 2023-07-17
CoT Explanation column in generated LLM outputs


nihit:
**Describe the bug**
Currently we don't export the explanation generated by the LLM when chain of thought flag is set to True

**Expected behavior**
We should support adding this in the "explanation column" similar to the seed csv.

nihit:
@turian -- this should be fixed now! thanks for flagging it

PR #462, opened 2023-07-17
Add llm explanation to output dataframe

rajasbansal:
Currently, the llm explanation is not exposed to the user. This PR adds explanations to LLMAnnotation and the output dataframe allowing users to see the explanation for a label if chain of thought is enabled

Address #460


nihit (approved):
lgtm!

PR #468, opened 2023-07-18
Fix metric calculation for unlabeled datapoints

### LINKED ISSUE #467: [Bug]: Metric calculation when the dataset is partially labeled
**Describe the bug**
If the dataset is partially labeled, i.e some rows have labels and the others do not, we still calculate metrics on the entire dataset.  This is need for metrics like completion rate, but for accuracy we should just consider just labeled datapoints.

**To Reproduce**
In any example dataset, remove the labels for 50% of the dataset and check the accuracy on this dataset.

**Expected behavior**
We should calculate accuracy only on the labeled examples and filter out the unlabeled examples while measuring accuracy.

rajasbansal:
Currently, we calculate metrics like accuracy and F1 on the entire dataset including unlabeled examples. This leads to incorrect numbers in the case the dataset has some labeled datapoints interspersed with unlabeled examples. This PR filters unlabeled examples before doing metric calculation.

Closes #467

### DIFF
--- a/src/autolabel/tasks/classification.py
+++ b/src/autolabel/tasks/classification.py
@@ -9,6 +9,7 @@ from autolabel.configs import AutolabelConfig
 from autolabel.schema import LLMAnnotation, Metric, MetricResult
 from autolabel.tasks import BaseTask
 from autolabel.utils import get_format_variables
+from autolabel.tasks.utils import filter_unlabeled_examples
 
 import json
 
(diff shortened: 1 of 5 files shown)


yadavsahil197 (approved):
LG!
We still see a few examples where the LLM doesn't conform to the expected output, those should be marked as unlabeled. We could have the model output a NULL TOKEN if it doesn't know the answer.
@nihit  and @rishabh-bhargava would appreciate your thoughts on this!

nihit (approved):

PR #470, opened 2023-07-19
added output cost calculation to PaLM get_cost

Tyrest:
Previously, we were not calculating the output costs of the PALM model. Added a default `max_output_tokens` param and output token calculation into `get_cost`.


rajasbansal (approved):
lgtm

PR #475, opened 2023-07-20
Bubble up errors and clean up error creation

rajasbansal:
- Errors have a structure with error type and error message
- Errors sent in addition to text generation
- New RefuelLLMResult which encapsulates text generations, errors and cost
- Confidence calculate now returns a float
- Errors saved along with annotations and saved in the dataframe

### DIFF
--- a/src/autolabel/models/refuel.py
+++ b/src/autolabel/models/refuel.py
@@ -9,6 +9,7 @@ from typing import List, Optional
 from autolabel.models import BaseModel
 from autolabel.configs import AutolabelConfig
 from autolabel.cache import BaseCache
+from autolabel.schema import LabelingError, ErrorType
 
 
 from tenacity import (
@@ -67,6 +68,7 @@ class RefuelLLM(BaseModel):
 
     def _label(self, prompts: List[str]) -> LLMResult:
         generations = []
+        errors = []
         for prompt in prompts:
             try:
                 if self.SEP_REPLACEMENT_TOKEN in prompt:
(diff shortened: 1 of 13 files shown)


src/autolabel/models/refuel.py
```diff
@@ -83,13 +85,17 @@ def _label(self, prompts: List[str]) -> LLMResult:
                     self.SEP_REPLACEMENT_TOKEN, "\n"
                 )
                 generations.append([Generation(text=response)])
+                errors.append(None)
             except Exception as e:
                 # This signifies an error in generating the response using RefuelLLm
                 logger.error(
                     f"Unable to generate prediction: {e}",
                 )
                 generations.append([Generation(text="")])
```
Tyrest, line 94: Should we append some type of error here?

Tyrest (approved):
lgtm!

src/autolabel/schema.py
```diff
@@ -71,6 +71,22 @@ class MetricResult(BaseModel):
     value: Any
 
 
+class ErrorType(str, Enum):
+    """Enum of supported error types"""
+
+    LLM_PROVIDER_ERROR = "llm_provider_error"
```
nihit, line 77: do we have a way to understand what is the LLM provider error ? e.g. rate limit / availability errors can be fixed with retries but context length errors are deterministic. 

if this is hard to parse and needs provider specific logic, okay to club into one for now but ideally we can bubble the specific error to the users.

src/autolabel/schema.py
```diff
@@ -172,3 +189,16 @@ class CacheEntry(BaseModel):
...
+    """List of generated outputs. This is a List[List[]] because
+    each input could have multiple candidate generations."""
+
+    generations: List[List[Generation]]
+
+    """Errors encountered while running the labeling job"""
+    errors: List[Optional[LabelingError]]
+
+    """Costs incurred during the labeling job"""
+    cost: Optional[float] = 0.0
```
nihit, line 204: Should we track cost individually as well? List[float] ? this is going to be needed for putting it in the row-level metadata anyway

nihit (approved):
lgtm otherwise!

src/autolabel/schema.py
rajasbansal: I think we should definitely handle context length errors differently. They are handled separately for different providers and there is no consistent langchain error that is sent. We can do better and combine this across providers. Might need some custom string matching on the error for different providers - tracking this as a separate ticket https://github.com/refuel-ai/autolabel/issues/481

PR #480, opened 2023-07-21
Metric calculation as a separate class

rajasbansal:
- This allows separation of metric calculation from all tasks
- Users can define new additional metrics as classes which can be run every time the model runs
- AUROC calculation now does not return thresholds
- Thresholding logic removed from all tasks - users can request the values at specific thresholds using the output object, or agent.report can show only the thresholds which the user should care about
- Allow specifying custom metrics

### DIFF
--- a/src/autolabel/metrics/base.py
+++ b/src/autolabel/metrics/base.py
@@ -1,0 +1,17 @@
+from abc import ABC, abstractmethod
+from typing import List
+
+from autolabel.schema import LLMAnnotation, MetricResult
+
+
+class BaseMetric(ABC):
+    value: List[MetricResult]
+
+    def __init__(self) -> None:
+        super().__init__()
+
+    @abstractmethod
+    def compute(
+        self, llm_labels: List[LLMAnnotation], gt_labels: List[str]
+    ) -> List[MetricResult]:
+        pass
--- a/src/autolabel/metrics/base.py
+++ b/src/autolabel/metrics/base.py
@@ -5,8 +5,6 @@ from autolabel.schema import LLMAnnotation, MetricResult
 
 
 class BaseMetric(ABC):
-    value: List[MetricResult]
-
     def __init__(self) -> None:
         super().__init__()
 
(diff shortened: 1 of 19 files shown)


src/autolabel/tasks/question_answering.py
```diff
@@ -106,85 +100,24 @@ def get_explanation_prompt(self, example: Dict) -> str:
         )
 
     def eval(
-        self, llm_labels: List[LLMAnnotation], gt_labels: List[str]
+        self,
+        llm_labels: List[LLMAnnotation],
+        gt_labels: List[str],
+        additional_metrics: Optional[List[BaseMetric]] = [],
```
rajasbansal, line 106: lmk if passing additional metrics over a set of default metrics for a task makes sense - we could have this be a set of metrics so you can also remove some default metrics

src/autolabel/metrics/base.py
```diff
@@ -0,0 +1,17 @@
+from abc import ABC, abstractmethod
+from typing import List
+
+from autolabel.schema import LLMAnnotation, MetricResult
+
+
+class BaseMetric(ABC):
+    value: List[MetricResult]
```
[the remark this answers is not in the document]
nihit: hmm, I think it's better to have this be stateless and return on compute if we're passing the inputs during the compute call - the intermediate way (inputs passed to compute, results stored in self.value) is a bit confusing?

nihit (commented):
still need to finish reviewing but initial thoughts

src/autolabel/tasks/question_answering.py
nihit: would a user specify additional metrics in the config ? (or, in the future world, initialize it in code themselves and pass it here) ?

rajasbansal: The hope is that these Metrics will be defined in code and the user can specify them in the agent.run command

rajasbansal: Added a test to show how a new metric can be added

nihit (approved):
lgtm!

PR #482, opened 2023-07-22
Output object instead of dataframe to handle sorting and filtering

rajasbansal:
1. Add an object as the result of run instead of a dataframe
2. The object contains the dataframe of labels as a field
3. Filter and eval according to the object

### DIFF
--- a/src/autolabel/labeling_output.py
+++ b/src/autolabel/labeling_output.py
@@ -101,7 +101,7 @@ class LabelingOutput:
         )
 
     def mistakes(self, label: str = None, ground_truth: str = None):
-        gt_label_column = self.config().label_column()
+        gt_label_column = self.config.label_column()
 
         if gt_label_column is None:
             raise ValueError(
@@ -125,7 +125,7 @@ class LabelingOutput:
         )
 
     def correct(self):
-        gt_label_column = self.config().label_column()
+        gt_label_column = self.config.label_column()
 
         if gt_label_column is None:
             raise ValueError("Cannot compute correct without ground truth label column")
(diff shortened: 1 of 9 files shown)


src/autolabel/labeling_output.py
```diff
@@ -0,0 +1,211 @@
...
+        if filter_func:
+            filtered_df = filtered_df.apply(filter_func, axis=1)
+
+        return LabelingOutput(
+            self.config,
+            filtered_df,
+            self.metrics,
+        )
+
+    def errors(self):
```
nihit, line 114: errors vs mistakes is a bit confusing naming - how about we roll it into the same function? 

llm label != gt label is ultimately a kind of error?

src/autolabel/labeling_output.py
```diff
@@ -0,0 +1,211 @@
...
+        filtered_df = self.df[self.df[self.prefix + "error"].notnull()]
+        return LabelingOutput(
+            self.config,
+            filtered_df,
+            self.metrics,
+        )
+
+    def non_errors(self):
```
nihit, line 122: can remove this for now

src/autolabel/labeling_output.py
```diff
@@ -0,0 +1,211 @@
...
+        filtered_df = self.df[
+            self.df[self.prefix + "label"] == self.df[gt_label_column]
+        ]
+        return LabelingOutput(
+            self.config,
+            filtered_df,
+            self.metrics,
+        )
+
+    def correct_and_confident(self, threshold: float = 0.5):
```
nihit, line 169: Suggest simplifying this to only filter by confidence (independent of correct)

nihit (approved):
lgtm otherwise!

src/autolabel/labeling_output.py
rajasbansal: I wanted to separate them out in order to find the examples where the model was not able to answer, if potentially they went over context length or some other input specific problem and we wanted to fix them

PR #485, opened 2023-07-25
Allow label column to be null for running autolabel

rajasbansal:
- Allow labeling of cases where the label column is None
- Fix NER bugs
- Allow the AutolabelConfig json schema to have nullable fields instead of strong data types


nihit (approved):
thank you for the quick fix!

nihit:
@rajasbansal good to merge this?

PR #493, opened 2023-07-27
cli file name bugfix

Tyrest:
The `autolabel init` would output the file as `[TODO] Enter task name_config.json` instead of `template_config.json`


Tyrest:
Also changed the drop down menu library from `enquiries` to `simple-term-menu` so we wouldn't have a dependency conflict (and since `enquiries` is no longer maintained)

Abhinav-Naikawadi (approved):

Issue #489, opened 2023-07-27
CLI wizard documentation


nihit:

Tyrest:
Doc [here](https://docs.google.com/document/d/1LAm25kBj6_nrXfBzzUa4NpKlHupFW4PZukjjCOCOjc0/edit?usp=sharing) to plan the looms demoing the CLI

PR #494, opened 2023-07-27
added CLI documentation

Tyrest:


docs/guide/resources/CLI.md
```diff
@@ -0,0 +1,99 @@
+The Autolabel CLI was created to make the config file creation process easier. It is a simple command line interface that will ask you a series of questions and then generate a config file for you. To use it, simply run the following command:
+
+```bash
+autolabel config
```
rajasbansal, line 4: can we walk through one example of the wizard interface?

docs/guide/resources/CLI.md
```diff
@@ -0,0 +1,99 @@
...
+```bash
+autolabel plan <path-to-dataset> <path-to-config>
+```
+
+## The `run` command
+The `run` command works identically to running `LabelingAgent({config}).run({dataset})` in python. To use it, simply run the following command:
+
+```bash
+autolabel run <path-to-dataset> <path-to-config>
+```
```
rajasbansal, line 99: it might not be possible to do this in the future (with the transform PR as you need to pass in the dataset to the agent function directly and the output is also a dataset) - not sure if we should keep this here. thoughts @Tyrest @nihit

Tyrest: After the transform PR, we could still make this command work for autolabel by reading the csv in the CLI code instead and likewise writing a new csv for output. I've been using `plan` quite a bit to just quickly what an example prompt might look like from a config and `run` might be useful for those who want to do a quick run for ~100 rows.

Tyrest:
#489

nihit (approved):

docs/guide/resources/CLI.md
nihit: Good callout, yes the suggested change makes sense to me - let's do this in a followup PR now that transforms related PR is merged

PR #495, opened 2023-07-27
added logit bias for huggingface

Tyrest:
Added logit bias for `huggingface_pipeline` models which can be enabled by settings `logit_bias` to `True` in the model config (just like `openai`)

### DIFF
--- a/src/autolabel/models/hf_pipeline.py
+++ b/src/autolabel/models/hf_pipeline.py
@@ -11,8 +11,6 @@ from autolabel.schema import RefuelLLMResult
 
 logger = logging.getLogger(__name__)
 
-DEFAULT_NUM_BEAMS = 4
-
 
 class HFPipelineLLM(BaseModel):
     DEFAULT_MODEL = "google/flan-t5-xxl"
@@ -51,7 +49,7 @@ class HFPipelineLLM(BaseModel):
         # populate model params
         model_params = config.model_params()
         self.model_params = {**self.DEFAULT_PARAMS, **model_params}
-        if config.logit_bias():
+        if config.logit_bias() != 0:
             self.model_params = {
                 **self._generate_sequence_bias(),
                 **self.model_params,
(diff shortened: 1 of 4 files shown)


src/autolabel/models/hf_pipeline.py
```diff
@@ -8,6 +9,11 @@
 from autolabel.schema import RefuelLLMResult
 
 
+logger = logging.getLogger(__name__)
+
+DEFAULT_NUM_BEAMS = 4
```
DhruvaBansal00, line 14: Is this constant being used anywhere? If not, let's remove it.

src/autolabel/models/hf_pipeline.py
```diff
@@ -84,6 +97,49 @@ def __init__(self, config: AutolabelConfig, cache: BaseCache = None) -> None:
...
+            raise ValueError(
+                "Could not import transformers python package. "
+                "Please it install it with `pip install transformers`."
+            )
+        tokenizer = AutoTokenizer.from_pretrained(
+            self.model_name, use_fast=False, add_prefix_space=True
+        )
+        sequence_bias = {
+            tuple([tokenizer.eos_token_id]): 100.0,
+            # tuple(tokenizer([label], add_special_tokens=False).input_ids[0]): float(
```
DhruvaBansal00, line 123: Let's remove all the commented code here

src/autolabel/models/hf_pipeline.py
```diff
@@ -84,6 +97,49 @@ def __init__(self, config: AutolabelConfig, cache: BaseCache = None) -> None:
...
+            # )
+            # for label in self.config.labels_list()
+        }
+        max_new_tokens = 0
+        for label in self.config.labels_list():
+            tokens = tuple(tokenizer([label], add_special_tokens=False).input_ids[0])
+            # sequence_bias[tokens] = 10.0
+            # sequence_bias[tuple([tokens[0]])] = 10.0
+            for token in tokens:
+                sequence_bias[tuple([token])] = 100.0
```
DhruvaBansal00, line 134: Let's make "100" a number that is configurable? Instead of storing a boolean, config.logit_bias() can be set to a float. And if its set to 0 (which should be default) then we won't apply logit_bias.

src/autolabel/models/hf_pipeline.py
```diff
@@ -84,6 +97,49 @@ def __init__(self, config: AutolabelConfig, cache: BaseCache = None) -> None:
...
+            max_new_tokens = max(max_new_tokens, len(tokens))
+
+        print(sequence_bias, max_new_tokens)
```
DhruvaBansal00, line 137: Let's remove this print statement too

Tyrest: Forgot I left that in there!

src/autolabel/models/hf_pipeline.py
Tyrest: Sounds good to me

src/autolabel/models/hf_pipeline.py
Tyrest: There's a num_beams parameter in huggingface that is generally recommended to be used with sequence biasing but it didn't make a difference in testing for me so I'll remove it

DhruvaBansal00 (approved):
Lgtm :)

PR #496, opened 2023-07-28
Add transform base class with a very simple agent.transform

rajasbansal:
- Adds a unified dataset class which is ingested and outputted by every agent function
- Add a BaseTransform class which every custom transform will implement
- Add transforms key to the config
- Allow registering custom transform
- Tested out a dummy custom transform


src/autolabel/dataset/dataset.py
```diff
@@ -0,0 +1,263 @@
...
+
+    class Config:
+        arbitrary_types_allowed = True
+
+    def __init__(
+        self,
+        dataset: Union[pd.DataFrame, str],
+        config: AutolabelConfig,
+        max_items: int = None,
+        start_index: int = 0,
```
nihit, line 34: i think these should still belong to run/plan/transform methods ? 

feels more like a runtime parameter rather than something needed to create the dataset itself - 

in practice, moving this to the dataset would mean the user needs to create a new dataset if they want to run labeling using the same config/base csv but max_items = 100, vs max_items = 1000

src/autolabel/labeler.py
```diff
@@ -58,10 +59,8 @@ def __init__(
 
     def run(
         self,
-        dataset: Union[str, pd.DataFrame],
-        max_items: Optional[int] = None,
+        dataset: AutolabelDataset,
         output_name: Optional[str] = None,
-        start_index: Optional[int] = 0,
```
nihit, line 64: see comment in AutolabelDataset init

nihit (approved):

PR #501, opened 2023-07-31
Accept config in LabelingAgent

rajasbansal:
Currently we can pass a dict or string to the Labeling Agent.

This PR will change the LabelingAgent to accept an AutolabelConfig object as an input (to be consistent with the input to the dataset object)

```
from autolabel import LabelingAgent, AutolabelConfig, AutolabelDataset
config = AutolabelConfig('config.json')
dataset = AutolabelDataset('test.csv', config)
agent = LabelingAgent(config)
```

### DIFF
--- a/src/autolabel/labeler.py
+++ b/src/autolabel/labeler.py
@@ -43,13 +43,13 @@ class LabelingAgent:
 
     def __init__(
         self,
-        config: Union[str, Dict],
+        config: AutolabelConfig,
         cache: Optional[bool] = True,
     ) -> None:
         self.db = StateManager()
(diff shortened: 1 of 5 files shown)


src/autolabel/labeler.py
```diff
@@ -43,13 +43,13 @@ class LabelingAgent:
 
     def __init__(
         self,
-        config: Union[str, Dict],
```
nihit, line 46: can we still accept these, at least until we update the docs and share the change with Discord? 

will be backwards incompatible otherwise and break all of our examples/docs

nihit (approved):
lgtm otherwise

Issue #499, opened 2023-07-31
Add Transform for PDF Files


Tyrest:
**Is your feature request related to a problem? Please describe.**
Users may want to load data from paths to pdf files specified in the dataset. Since this describes a fairly common transform, autolabel should support it natively.

**Describe the solution you'd like**
When a user uses paths to pdfs as input, we should be able to read the contents of the pdf file and provide it as input for the labeling agent.

**Additional context**
Sub issue of #403

Tyrest:
Could add OCR with one of [docquery](https://github.com/impira/docquery/blob/main/src/docquery/document.py), [pdfplumber](https://github.com/jsvine/pdfplumber), or [tesseract](https://github.com/tesseract-ocr/tesseract).

A good dataset for OCR testing: [image pdf dataset](https://huggingface.co/datasets/xia111/pdf/viewer/xia111--pdf/train)

PR #502, opened 2023-08-01
added pdf transform and tests

Tyrest:
#499


src/autolabel/transforms/__init__.py
```diff
@@ -1,12 +1,13 @@
...
 
 from .base import BaseTransform
+from .pdf import PDFTransform
 from typing import Dict, List
 from autolabel.schema import TransformType
 
 logger = logging.getLogger(__name__)
 
-TRANSFORM_REGISTRY = {}
+TRANSFORM_REGISTRY = {PDFTransform.name(): PDFTransform}
```
rajasbansal, line 10: Let's move this to schema.py as a TransformType

rajasbansal:
We will need pyproject.toml changes, we would need to add pypdf as a dependency

pyproject.toml
```diff
@@ -90,7 +90,8 @@ all = [
     "transformers >= 4.25.0",
     "google-cloud-aiplatform>=1.25.0",
     "cohere>=4.11.2",
-    "sentence_transformers"
+    "sentence_transformers",
+    "pypdf >= 3.14.0"
```
nihit, line 94: lets make this an optional dependency

src/autolabel/transforms/pdf.py
```diff
@@ -0,0 +1,68 @@
...
+        Returns:
+            Dict[str, any]: The transformed row of data.
+        """
+        try:
+            from langchain.document_loaders import PyPDFLoader
+        except ImportError:
+            raise ImportError(
+                "pypdf is required to use the pdf transform. Please install pypdf with the following command: pip install pypdf"
+            )
+        pages = self.extract_text(row[self.file_path_column])
```
nihit, line 60: hmm, I'm trying to understand how the langchain pypdf loader and tesseract OCR. the ideal behavior imo is to make native pdf document loading the default, and go the OCR route only if a "ocr_enabled" flag is turned on

rajasbansal: +1

Also, lets try to import on transform init instead of the .transform method so that once a transform object is initialized we know that we have the necessary things to run the transform

src/autolabel/transforms/pdf.py
```diff
@@ -0,0 +1,68 @@
+from typing import List, Dict, Any
+
+from pdf2image import convert_from_path
```
rajasbansal, line 3: do these packages need to be installed separately as well?

rajasbansal (changes requested):

src/autolabel/transforms/pdf.py
```diff
@@ -0,0 +1,68 @@
...
+        Each page is then converted into text and appended to the output string.
+
+        Args:
+            row (Dict[str, any]): The row of data to be transformed.
+
+        Returns:
+            Dict[str, any]: The transformed row of data.
+        """
+        try:
+            from langchain.document_loaders import PyPDFLoader
```
rajasbansal, line 55: Where are we using the PyPDFLoader?

src/autolabel/transforms/pdf.py
Tyrest: Yep, just moved them

src/autolabel/transforms/pdf.py
Tyrest: Thanks for catching this, just a leftover from previous iterations

src/autolabel/transforms/pdf.py
```diff
@@ -0,0 +1,102 @@
...
+        page_sep: str = "\n\n",
+    ) -> None:
+        """The output columns for this class should be in the order: [content_column, num_pages_column]"""
+        super().__init__(output_columns)
+        self.file_path_column = file_path_column
+        self.ocr_enabled = ocr_enabled
+        self.page_header = page_header
+        self.page_sep = page_sep
+
+        if self.ocr_enabled:
```
rajasbansal, line 22: If these are relatively disjoint considering if these should be put in separate classes as two separate PDFTransforms. This could make sense if we are planning to add more kinds of things like table reading to the pdf transform

@nihit @yadavsahil197 what do you think

rajasbansal (commented):
@Tyrest lgtm. Lets discuss today with @yadavsahil197  to coordinate a change to the interface (adding async support for transforms). We can add that to this PR and then should be good to land

src/autolabel/transforms/pdf.py
nihit: hmm, i would vote to keep it in the same one - easier for users to think of one transform for anything related to PDF loading - wdyt ?

rajasbansal: sg

rajasbansal (approved):

Issue #506, opened 2023-08-04
Add support for Webpage Transform


yadavsahil197:
Webpage Transform supports loading webpages and extracting text from provided urls.

nihit:
@rajasbansal  @nihit for review

PR #507, opened 2023-08-04
Add support for Webpage Transform

yadavsahil197:
Adds #506


src/autolabel/labeler.py
```diff
@@ -334,20 +339,29 @@ def plan(
...
+            transform.apply(input_dict) for input_dict in dataset.inputs
+        ]
+        outputs = await gather_async_tasks_with_progress(
+            transform_outputs,
+            description=f"Running transform {transform.name()}...",
+            console=console,
+        )
+        output_df = pd.DataFrame.from_records(outputs)
+        final_df = pd.concat([dataset.df, output_df], axis=1)
+        dataset = AutolabelDataset(final_df, self.config)
```
rajasbansal, line 355: we should probably add this as a util function in autolabelDataset so we can change this if needed in the future. I missed this when creating the base transform

src/autolabel/transforms/webpage_transform.py
```diff
@@ -0,0 +1,115 @@
...
+                timeout=self.timeout, limits=limits, follow_redirects=True, verify=False
+            )
+            self.beautiful_soup = BeautifulSoup
+        except ImportError:
+            raise ImportError(
+                "BeautifulSoup, httpx and fake_useragent are required to use the webpage transform. Please install them with the following command: pip install bs4 httpx fake_useragent"
+            )
+
+    def name(self) -> str:
+        return TransformType.WEBPAGE_TRANSFORM
```
rajasbansal, line 60: consider if adding parameters to the name makes sense so that the name of a transform is more unique

src/autolabel/transforms/webpage_transform.py
```diff
@@ -0,0 +1,115 @@
...
+    def _load_metadata(self, url, soup) -> Dict[str, Any]:
+        metadata = {"url": url}
+        if soup.find("title"):
+            metadata["title"] = soup.find("title").get_text()
+        for meta in soup.find_all("meta"):
+            if meta.get("name") and meta.get("content"):
+                metadata[meta.get("name")] = meta.get("content")
+            elif meta.get("property") and meta.get("content"):
+                metadata[meta.get("property")] = meta.get("content")
+        return metadata
```
rajasbansal, line 71: metadata will be missing some columns in case we dont find them in the page right? How will this be handled by _apply?

src/autolabel/transforms/webpage_transform.py
```diff
@@ -0,0 +1,115 @@
...
+            return None, None
+
+        try:
+            client = self.client
+            if not verify:
+                client = self.client_with_no_verify
+            response = await client.get(url, headers=headers)
+
+            # TODO: Add support for other parsers
+            soup = self.beautiful_soup(response.text, HTML_PARSER)
```
rajasbansal, line 87: It make sense to add parsers as part of the config and then the init of the transform will handle instantiating the parser.

However, in the future we might want to have a parsing class which can parse the webpage as you want and we define the usual parsers, i.e text parser, url parser, table parser etc. In that case even this might be passed as a class. However, that is difficult to do in the config ways of doing things and registering parsers seems to be out of scope for now

Edit: (Adding parsing as a separate transform is also possible however we need html in order to transform, extracting urls may not be possible with the text format)

src/autolabel/utils.py
```diff
@@ -158,6 +158,55 @@ def track(
...
+            progress_task,
+            advance=min(advance, total - progress.tasks[progress_task].completed),
+        )
+        progress.refresh()
+        return res
+
+    with progress:
+        progress_task = progress.add_task(description, total=total)
+        tasks = [_task_with_tracker(task, progress, progress_task) for task in tasks]
+        import asyncio
```
rajasbansal, line 205: any reason for importing inside this?

pyproject.toml
```diff
@@ -90,7 +92,10 @@ all = [
     "transformers >= 4.25.0",
     "google-cloud-aiplatform>=1.25.0",
     "cohere>=4.11.2",
-    "sentence_transformers"
+    "sentence_transformers",
+    "bs4",
+    "httpx",
+    "fake_useragent"
```
rajasbansal, line 98: do we need to add asyncio to the pyproject file?

rajasbansal (commented):
lgtm

rajasbansal (approved):

src/autolabel/transforms/webpage_transform.py
yadavsahil197: Right now, we are returning all the meta fields that are present from the webpage. An alternate could be to always look for specific fields, such as the description, encoding, social media tags etc.

pyproject.toml
yadavsahil197: think it is a built-in module

src/autolabel/utils.py
yadavsahil197: good catch! Can move it to the top.

src/autolabel/transforms/webpage_transform.py
yadavsahil197: I agree we can keep them as two separate set of transforms. The webpage transform just loads a given url and returns the loaded html, text or soup. The parser in this context refers to the beautiful soup parsers such as html.parser, lxml and html5lib.
The other transform support extracting content/urls from a raw html, bytes or soup.
That way each of these transforms have separation of concerns and be extended independent of each other.

src/autolabel/transforms/webpage_transform.py
yadavsahil197: Do you have any specific parameters in mind?

src/autolabel/transforms/webpage_transform.py
rajasbansal: Makes sense. In case one webpage has a title but the other webpage doesn't have a title we will put nan for the second page under the webpage column. This is being handled by pandas. Tried it out using 

```
import pandas as pd
l = [{'a': 1, 'b': 2}, {'a': 3, 'b': 4}]
pd.DataFrame(l)
   a    b    c
0  1  2.0  NaN
1  3  NaN  4.0
```

rajasbansal (approved):

PR #508, opened 2023-08-08
Add cache for transforms and unpin sqlite

### LINKED ISSUE #138: Update sqlalchemy to latest version
This will require us to remove usage of autocommit=True when initializing engine. So we'll need to explicitly commit a change/transaction.

rajasbansal:
- Add caching support for transforms
- Remove autocommit from sqlite as this breaks under async operations
- Unpin sqlite as we no longer use autocommit

Closes #138 #510


src/autolabel/cache/sqlalchemy_transform_cache.py
```diff
@@ -0,0 +1,51 @@
...
+from typing import Optional
+from autolabel.data_models import TransformCacheEntryModel
+from .base import BaseCache
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class SQLAlchemyTransformCache(BaseCache):
+    """A cache system implemented with SQL Alchemy"""
```
yadavsahil197, line 15: probably wanna add more information here

src/autolabel/data_models/dataset.py
```diff
@@ -21,6 +21,7 @@ def __repr__(self):
     def create(cls, db, dataset: BaseModel):
         db_object = cls(**json.loads(dataset.json()))
         db.add(db_object)
+        db.commit()
```
yadavsahil197, line 24: we should also commit for delete

src/autolabel/data_models/task_run.py
```diff
@@ -40,7 +40,7 @@ def create(cls, db, task_run: BaseModel):
         logger.debug(f"creating new task: {task_run}")
         db_object = cls(**task_run.dict())
         db.add(db_object)
-        db.flush()
+        db.commit()
```
yadavsahil197, line 43: we should also commit on update?

src/autolabel/labeler.py
```diff
@@ -478,8 +483,15 @@ def generate_explanations(
...
-        else:
-            logger.error("No cache to clear")
+    def clear_cache(self, use_ttl: bool = True):
+        """
+        Clears the generation and transformation cache from autolabel.
+        Args:
+            use_ttl: If true, only clears the cache if the ttl has expired.
+            If this is true, only the transform cache will be cleared (as the generation cache does not use ttl)
+        """
+        if self.generation_cache and not use_ttl:
```
yadavsahil197, line 493: should we make the default ttl to be a large number so we don't need to handle it any different than the transform cache?

yadavsahil197 (approved):
LGTM otherwise

src/autolabel/labeler.py
rajasbansal: makes sense, done

pyproject.toml
```diff
@@ -27,7 +27,7 @@ dependencies = [
     "pandas >= 1.3.0",
     "scikit-learn >= 1.0.0",
     "tenacity >= 8.2.2",
-    "SQLAlchemy == 1.4.47",
```
nihit, line 30: nice!

src/autolabel/transforms/pdf.py
```diff
@@ -2,19 +2,21 @@
 
 from autolabel.schema import TransformType
 from autolabel.transforms import BaseTransform
+from autolabel.cache import BaseCache
 
 
 class PDFTransform(BaseTransform):
     def __init__(
         self,
+        cache: BaseCache,
```
nihit, line 11: in the BaseTransform init `cache` is the last param, we should probably maintain this for other downstream transforms as well - not a requirement but for readability

src/autolabel/transforms/webpage_transform.py
```diff
@@ -19,12 +21,13 @@
 class WebpageTransform(BaseTransform):
     def __init__(
         self,
+        cache: BaseCache,
```
nihit, line 24: same

src/autolabel/cache/sqlalchemy_transform_cache.py
```diff
@@ -0,0 +1,56 @@
...
+            entry: TransformCacheEntry we wish to retrieve from the Cache
+        Returns:
+            result: The output of the transform for this input. None if not found.
+        """
+        cache_entry = TransformCacheEntryModel.get(self.session, entry)
+        if cache_entry is None:
+            logger.debug("Cache miss")
+            return None
+
+        logger.debug("Cache hit")
```
nihit, line 44: remove the log statements?

nihit (commented):
looks good!

PR #515, opened 2023-08-09
added ocr image transform

Tyrest:
Added image transform capable of reading text from the image types listed below:
- PNG
- JPEG
- TIFF
- JPEG 2000
- GIF
- WebP
- BMP
- PNM


.gitignore
```diff
@@ -143,4 +143,7 @@ examples/*/*.csv
 *.pkl
 
 # DB Files
-*.db
\ No newline at end of file
+*.db
+
+# Image Files
+*.png
```
rajasbansal, line 149: will this lead to png files in docs being ignored?

rajasbansal (approved):

.gitignore
Tyrest: oh that's a good point I'll change that

PR #517, opened 2023-08-10
Update docs and add an example webpage transform notebook

rajasbansal:
- Update the docs for autolabel dataset and transforms
- Add an example notebook for running webpage transforms
- Serialize llm annotations as pickle dumps as they contain objects in the curr_sample (input row) which is not serializable to store in the cache

Open to suggestions on the serialization @nihit


src/autolabel/transforms/webpage_transform.py
```diff
@@ -102,7 +102,7 @@ async def _load_url(
             soup = self.beautiful_soup(content_bytes, HTML_PARSER)
             return {
                 "content": soup.get_text(),
-                "content_bytes": content_bytes,
+                # "content_bytes": content_bytes,
```
nihit, line 105: let's just remove this ?

src/autolabel/labeler.py
```diff
@@ -31,6 +32,7 @@
     track,
     track_with_stats,
     gather_async_tasks_with_progress,
+    dump_strings_in_dict,
```
nihit, line 35: is this being used anywhere?

src/autolabel/tasks/base.py
```diff
@@ -16,7 +17,11 @@
     LabelingError,
     ErrorType,
 )
-from autolabel.utils import get_format_variables, extract_valid_json_substring
+from autolabel.utils import (
+    get_format_variables,
+    extract_valid_json_substring,
+    dump_strings_in_dict,
```
nihit, line 23: same

nihit (approved):
lgtm otherwise!

PR #531, opened 2023-08-11
Fix AUROC Bug and allow custom Example Selectors

rajasbansal:
- Allow passing custom example selectors which extend the BaseExampleSelector interface
- Fix AUROC when only class is present


Abhinav-Naikawadi (commented):
I think all the docs/examples are accidental right?

rajasbansal:
updated with correct changes

Abhinav-Naikawadi (approved):

PR #540, opened 2023-08-15
Transforms: Add error handling

### LINKED ISSUE #513: Support error handling in transforms

yadavsahil197:
Fixes #513

### DIFF
--- a/src/autolabel/transforms/base.py
+++ b/src/autolabel/transforms/base.py
@@ -1,7 +1,10 @@
 from abc import ABC, abstractmethod
-from typing import Dict, Any
 from autolabel.cache import BaseCache
 from autolabel.schema import TransformCacheEntry
+from typing import Dict, Any
+import logging
+
+logger = logging.getLogger(__name__)
 
 
 class BaseTransform(ABC):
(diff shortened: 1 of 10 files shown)


src/autolabel/transforms/base.py
```diff
@@ -71,7 +81,14 @@ async def apply(self, row: Dict[str, Any]) -> Dict[str, Any]:
                 # Cache hit
                 return output
 
-        output = await self._apply(row)
+        try:
+            output = await self._apply(row)
+            output[self.transform_status_column] = True
+        except Exception as e:
+            logger.error(f"Error applying transform {self.name()}. Exception: {e}")
+            output = {k: None for k in self.output_columns.values() if k is not None}
```
rajasbansal, line 89: Do we want to put this value as something standard like the NULL_TRANSFORM_TOKEN. Some string value instead of none might make sense

src/autolabel/transforms/base.py
```diff
@@ -71,7 +81,14 @@ async def apply(self, row: Dict[str, Any]) -> Dict[str, Any]:
                 # Cache hit
                 return output
 
-        output = await self._apply(row)
+        try:
+            output = await self._apply(row)
+            output[self.transform_status_column] = True
+        except Exception as e:
+            logger.error(f"Error applying transform {self.name()}. Exception: {e}")
```
rajasbansal, line 88: can we put the error message in applied successfully as well? Would be easier to know what the error was for a specific row along with seeing the logs.

rajasbansal, line 88: for errors consider if error codes make sense, to keep this aligned with how errors are handled for generations

rajasbansal (approved):
lgtm, left some comments on how errors are stored

src/autolabel/transforms/base.py
yadavsahil197: I think we should not store a boolean and string value in the same column. To store the error we might go with just transformation_error as a column instead of applied successfully. WDYT

rajasbansal: makes sense

PR #543, opened 2023-08-16
Fix error message on openai key not set

rajasbansal:


Abhinav-Naikawadi (approved):

src/autolabel/models/__init__.py
```diff
@@ -50,10 +50,13 @@ def from_config(config: AutolabelConfig, cache: BaseCache = None) -> BaseModel:
             assert isinstance(
                 model_obj, BaseModel
             ), f"{model_obj} should inherit from autolabel.models.BaseModel"
-        except ValueError as e:
+        except KeyError as e:
+            # We should never get here as the config should have already
+            # been validated by the pydantic model.
             logger.error(
```
Abhinav-Naikawadi, line 56: we should change this log right?

Tyrest (approved):

src/autolabel/models/__init__.py
rajasbansal: Nope in case there is a KeyError i.e the dictionary does not have the provider name we will see this error

PR #544, opened 2023-08-16
added docs for synthetic datasets and progress

Tyrest:
Added documentation for synthetic dataset generation and also added a progress bar to indicate how long generating the dataset will take.


docs/guide/resources/synthetic_dataset_generation.md
```diff
@@ -0,0 +1,65 @@
...
+  },
+  "model": {
+    "provider": "openai",
+    "name": "gpt-3.5-turbo"
+  },
+  "prompt": {
+    "task_guidelines": "You are an expert at understanding bank customers support complaints and queries.\nYour job is to correctly classify the provided input example into one of the following categories.\nCategories:\n{labels}",
+    "output_guidelines": "You will answer with just the the correct output label and nothing else.",
+    "labels": [
+      "activate_my_card",
```
rajasbansal, line 33: how are we adding label descriptions

docs/guide/resources/synthetic_dataset_generation.md
```diff
@@ -0,0 +1,65 @@
...
+      ..., # other labels
+      "wrong_exchange_rate_for_cash_withdrawal"
+    ],
+    "example_template": "Input: {example}\nOutput: {label}"
+  },
+  "dataset_generation": {
+    "num_rows": 5,
+    "guidelines": "You are an expert at generating synthetic data. You will generate a dataset that satisfies the following criteria:\n1. The data should be diverse and cover a wide range of scenarios.\n2. The data should be as realistic as possible, closely mimicking real-world data.\n3. The data should vary in length, some shorter and some longer.\n4. The data should be generated in a csv format.\n\nEach row should contain a realistic bank complaint. Use CSV format, with each line containing just the complaint and nothing else."
```
rajasbansal, line 41: Do we give the model an example input? eg. few shot?

nihit:
same question as Rajas on few shot in the config. else looks good

docs/guide/resources/synthetic_dataset_generation.md
Tyrest: I thought we decided to not use few shot because if they had those examples already, they wouldn't need to generate more. I can add it back in a following PR though!

docs/guide/resources/synthetic_dataset_generation.md
Tyrest: We aren't in this case, which would just provide the labels to the LLM without descriptions. I did forget to add it to the docs though so thanks for reminding me!

docs/guide/resources/synthetic_dataset_generation.md
rajasbansal: hmm makes sense. I was under the impression we send some example input to get good results. If that's not true, no few shot is also fine. However, in that case, how do we specify the type of each input and what each input means?

rajasbansal: update: read the answer for the last question, looks like it is going to be added to the description. Sounds good in that case

docs/guide/resources/synthetic_dataset_generation.md
```diff
@@ -43,6 +43,17 @@ config = {
 }
 ```
 
+If you want to provide label descriptions to further help the LLM understand what each label means, you can change the label list to a dictionary as shown below:
+
+```python
```
rajasbansal, line 48: lets add this to the default way of doing dataset generation because I think this would be important to get reasonable results to start with

nihit (approved):

PR #545, opened 2023-08-17
fix bugs for kill signal and indexing

Tyrest:
If the `run` method was stopped by kill signal while a progress bar was being displayed, the next time the `run` method would be run would result in the error `LiveError: Only one live display may be active at once`. This was fixed by setting self.console to Console(quiet=not console_output) instead of writing an if statement whenever printing output.

Additionally, when generating synthetic datasets, the index would repeat for every label (0, 1, 2, 0, 1, 2, ...) when we simply want the index to be a sequence from 0 to n-1. This was fixed by setting `ignore_index` to `True` when concatenating dataframes. A progress bar was also added for generating synthetic datasets.

### DIFF
--- a/src/autolabel/labeler.py
+++ b/src/autolabel/labeler.py
@@ -61,7 +61,7 @@ class LabelingAgent:
         self.db = StateManager() if self.create_task else None
         self.generation_cache = SQLAlchemyGenerationCache() if cache else None
         self.transform_cache = SQLAlchemyTransformCache() if cache else None
-        self.console = Console() if console_output else None
+        self.console = Console() if console_output else Console(quiet=True)
 
         self.config = (
             config if isinstance(config, AutolabelConfig) else AutolabelConfig(config)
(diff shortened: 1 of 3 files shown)


Tyrest:
Had to change utils too since there was a problem with not actually using the console that was provided to `track_with_stats`

src/autolabel/labeler.py
```diff
@@ -170,17 +170,12 @@ def run(
 
         indices = range(current_index, len(dataset.inputs))
 
-        if self.console:
```
rajasbansal, line 173: We need the ability to not print to the console. @nihit had earlier removed this for use in the cloud product, want to confirm with him if we should remvoe this

Tyrest: I didn't remove the feature, rich has a built in parameter to pass to the console when creating it called `quiet` which suppresses all output. I set this to `not console_output` in the constructor instead of using if statements everywhere since I think it's a cleaner solution (and also resolves the kill signal bug).

PR #546, opened 2023-08-17
[Draft] Attribute extraction

Tyrest:
Adds a new task called `attribute_extraction` to the supported tasks. This task aims to extract multiple attributes from text and return it as a dictionary. It requires a new section in the config called `output_attributes` which details the `name` and `options` for each attribute. This could be expanded further to support more than just classification style labels such as multilabel classification or question answering.

Currently the accuracy and f1 metrics are calculated individually for every attribute. This could be expanded by specifying what type each attribute is in the `output_attributes` dictionary and deciding which evals to compute for each attribute based on that.

Here's an example of an eval table for the supreme court dataset:
Screenshot 2023-08-17 at 3 07 07 PM


Tyrest:
With the most recent commit, the user can now specify what type the attribute is (currently supports classification, multilabel, and question answering) which allows autolabel to compute the correct labels for the correct attributes.

rajasbansal:
Working on top of this PR. Thanks a lot @Tyrest!

PR #547, opened 2023-08-22
Add support for Serp Api Transform

yadavsahil197:


src/autolabel/transforms/serp_api.py
```diff
@@ -0,0 +1,80 @@
...
+        try:
+            search_result = await self.serp_api_wrapper.arun(query=query)
+        except Exception as e:
+            raise TransformError(
+                TransformErrorType.SERP_API_ERROR,
+                f"Error while making request to Serp API: {e}",
+            )
+        return search_result
+
+    async def _apply(self, row: Dict[str, Any]) -> Dict[str, Any]:
```
rajasbansal, line 62: There is some preprocessing done by the Langchain SerpAPI response which is useful (for eg - reading the answerBox if it is present, then read the side bar and so on...)
Is it useful for us to do this as well (I found it useful while doing Enigma's smaller tasks as the answerbox is correct with a higher accuracy than other things)

https://github.com/langchain-ai/langchain/blob/acb54d8b9da9c5bcf73fade48c3c20f54388662a/libs/langchain/langchain/utilities/serpapi.py#L128

Do we want to bring this into our transform so that we can edit it or do you think that the response from Langchain is good enough?

src/autolabel/transforms/serp_api.py
```diff
@@ -0,0 +1,80 @@
...
+        query = row[self.query_column]
+        search_result = {}
+        if pd.isna(query):
```
rajasbansal, line 65: Do we also want to check for the null transform key?

yadavsahil197: The SerpAPIWrapper will throw an error for null serp API key if that's what you are referring to.

src/autolabel/transforms/serp_api.py
rajasbansal: While it is possible to do processing on the raw json response by the user, we should also provide some opinionated ways of processing the json so that the response can be used by the LLM more usefully. We can fill this in as we learn about tasks that use this transform. Hard to figure out what will work and what won't without experimentation.

Can add this as a follow up when we learn more

src/autolabel/transforms/serp_api.py
rajasbansal: I wanted to say if the input element is NULL_TRANSFORM_LABEL which you get as a result of an error with a transform

rajasbansal (approved):

src/autolabel/transforms/serp_api.py
yadavsahil197: Makes sense, adding handling for the null transform token

Issue #554, opened 2023-08-26
[Bug]: SemanticSimilarityExampleSelector error when trying to embed non-text columns from a dataset


nihit:
From Discord:

It seems like the few_shot learning examples can't handle types other than strings or that the data when read in  from the CSV file is not handling non-string values

In the method from_examples in class SemanticSimilarityExampleSelector in the semantic_similarity.py file, the code breaks on line
```
string_example = " ".join(
    sorted_values({key: example[key] for key in self.input_keys})
)
```

After putting some print statements in the method. It seems that it was breaking because one of my columns had float values and the another column had bool values. 
Error

```
    query = " ".join(sorted_values(input_variables))
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: sequence item 0: expected str instance, bool found
```

PR #558, opened 2023-08-29
Fix example selector with non string inputs

rajasbansal:
Addresses #554 
Changes the input variables to strings when possible and changes to empty strings when not possible

### DIFF
--- a/src/autolabel/utils.py
+++ b/src/autolabel/utils.py
@@ -383,3 +383,14 @@ def in_notebook():
     except AttributeError:
         return False
     return True
+
+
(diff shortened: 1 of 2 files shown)


src/autolabel/utils.py
```diff
@@ -383,3 +383,14 @@ def in_notebook():
     except AttributeError:
         return False
     return True
+
+
+def try_convert_values_to_strings(data: Dict) -> Dict:
```
nihit, line 388: nit: "safe_serialize_to_string"

nihit (approved):
looks good otherwise

PR #561, opened 2023-09-01
Better support for classification tasks with large number of label classes

iomap:
New option to filter labels_list by similarity to input example.

Two new optional fields are now present in prompt_config schema:
```
        "label_selection": true,
        "label_selection_count": 10
```

### DIFF
--- a/src/autolabel/labeler.py
+++ b/src/autolabel/labeler.py
@@ -174,10 +174,10 @@ class LabelingAgent:
         if self.config.label_selection():
             self.label_selector = LabelSelector.from_examples(
                 labels=self.config.labels_list(),
-                k=self.config.label_selection_count(),
                 embedding_func=PROVIDER_TO_MODEL.get(
                     self.config.embedding_provider(), DEFAULT_EMBEDDING_PROVIDER
                 )(),
+                k=self.config.label_selection_count(),
             )
 
         current_index = self.task_run.current_index if self.create_task else 0
@@ -356,10 +356,10 @@ class LabelingAgent:
         if self.config.label_selection():
             self.label_selector = LabelSelector.from_examples(
                 labels=self.config.labels_list(),
-                k=self.config.label_selection_count(),
                 embedding_func=PROVIDER_TO_MODEL.get(
                     self.config.embedding_provider(), DEFAULT_EMBEDDING_PROVIDER
                 )(),
+                k=self.config.label_selection_count(),
             )
 
         input_limit = min(len(dataset.inputs), 100)
--- a/src/autolabel/labeler.py
+++ b/src/autolabel/labeler.py
@@ -17,6 +17,7 @@ from autolabel.dataset import AutolabelDataset
 from autolabel.data_models import AnnotationModel, TaskRunModel
 from autolabel.database import StateManager
 from autolabel.few_shot import ExampleSelectorFactory, BaseExampleSelector
+from autolabel.few_shot.label_selector import LabelSelector
 from autolabel.models import BaseModel, ModelFactory
 from autolabel.metrics import BaseMetric
 from autolabel.transforms import BaseTransform, TransformFactory
@@ -165,6 +166,11 @@ class LabelingAgent:
                 cache=self.generation_cache is not None,
             )
 
+        if self.config.label_selection():
+            self.label_selector = LabelSelector.from_examples(
+                labels=self.config.labels_list()
+            )
+
         current_index = self.task_run.current_index if self.create_task else 0
         cost = 0.0
         postfix_dict = {}
@@ -185,8 +191,14 @@ class LabelingAgent:
                 )
             else:
                 examples = []
-                # Construct Prompt to pass to LLM
-            final_prompt = self.task.construct_prompt(chunk, examples)
+            # Construct Prompt to pass to LLM
+            if self.config.label_selection():
+                selected_labels = self.label_selector.select_labels(chunk["example"])
+                final_prompt = self.task.construct_prompt(
+                    chunk, examples, selected_labels
+                )
+            else:
+                final_prompt = self.task.construct_prompt(chunk, examples)
 
             response = self.llm.label([final_prompt])
             for i, generations, error in zip(
@@ -332,6 +344,11 @@ class LabelingAgent:
             cache=self.generation_cache is not None,
         )
 
+        if self.config.label_selection():
+            self.label_selector = LabelSelector.from_examples(
+                labels=self.config.labels_list()
+            )
+
         input_limit = min(len(dataset.inputs), 100)
 
         for input_i in track(
@@ -346,7 +363,13 @@ class LabelingAgent:
                 )
             else:
                 examples = []
-            final_prompt = self.task.construct_prompt(input_i, examples)
+            if self.config.label_selection():
+                selected_labels = self.label_selector.select_labels(input_i["example"])
+                final_prompt = self.task.construct_prompt(
+                    input_i, examples, selected_labels
+                )
+            else:
+                final_prompt = self.task.construct_prompt(input_i, examples)
             prompt_list.append(final_prompt)
 
             # Calculate the number of tokens
--- a/src/autolabel/labeler.py
+++ b/src/autolabel/labeler.py
@@ -16,7 +16,12 @@ from autolabel.configs import AutolabelConfig
 from autolabel.dataset import AutolabelDataset
 from autolabel.data_models import AnnotationModel, TaskRunModel
 from autolabel.database import StateManager
-from autolabel.few_shot import ExampleSelectorFactory, BaseExampleSelector
+from autolabel.few_shot import (
+    ExampleSelectorFactory,
+    BaseExampleSelector,
+    DEFAULT_EMBEDDING_PROVIDER,
+    PROVIDER_TO_MODEL,
+)
 from autolabel.few_shot.label_selector import LabelSelector
 from autolabel.models import BaseModel, ModelFactory
 from autolabel.metrics import BaseMetric
@@ -168,7 +173,11 @@ class LabelingAgent:
 
         if self.config.label_selection():
             self.label_selector = LabelSelector.from_examples(
-                labels=self.config.labels_list()
+                labels=self.config.labels_list(),
+                k=self.config.label_selection_count(),
+                embedding_func=PROVIDER_TO_MODEL.get(
+                    self.config.embedding_provider(), DEFAULT_EMBEDDING_PROVIDER
+                )(),
             )
 
         current_index = self.task_run.current_index if self.create_task else 0
@@ -346,7 +355,11 @@ class LabelingAgent:
 
         if self.config.label_selection():
             self.label_selector = LabelSelector.from_examples(
-                labels=self.config.labels_list()
+                labels=self.config.labels_list(),
+                k=self.config.label_selection_count(),
+                embedding_func=PROVIDER_TO_MODEL.get(
+                    self.config.embedding_provider(), DEFAULT_EMBEDDING_PROVIDER
+                )(),
             )
 
         input_limit = min(len(dataset.inputs), 100)
--- a/src/autolabel/labeler.py
+++ b/src/autolabel/labeler.py
@@ -31,6 +31,7 @@ from autolabel.schema import (
     MetricResult,
     TaskRun,
     TaskStatus,
+    TaskType,
 )
 from autolabel.tasks import TaskFactory
 from autolabel.utils import (
@@ -172,13 +173,18 @@ class LabelingAgent:
             )
 
         if self.config.label_selection():
-            self.label_selector = LabelSelector.from_examples(
-                labels=self.config.labels_list(),
-                embedding_func=PROVIDER_TO_MODEL.get(
-                    self.config.embedding_provider(), DEFAULT_EMBEDDING_PROVIDER
-                )(),
-                k=self.config.label_selection_count(),
-            )
+            if self.config.task_type() != TaskType.CLASSIFICATION:
+                self.console.print(
+                    "Warning: label_selection only supported for classification tasks!"
+                )
+            else:
+                self.label_selector = LabelSelector.from_examples(
+                    labels=self.config.labels_list(),
+                    embedding_func=PROVIDER_TO_MODEL.get(
+                        self.config.embedding_provider(), DEFAULT_EMBEDDING_PROVIDER
+                    )(),
+                    k=self.config.label_selection_count(),
+                )
 
         current_index = self.task_run.current_index if self.create_task else 0
         cost = 0.0
@@ -201,7 +207,10 @@ class LabelingAgent:
             else:
                 examples = []
             # Construct Prompt to pass to LLM
-            if self.config.label_selection():
+            if (
+                self.config.label_selection()
+                and self.config.task_type() == TaskType.CLASSIFICATION
+            ):
                 selected_labels = self.label_selector.select_labels(chunk["example"])
                 final_prompt = self.task.construct_prompt(
                     chunk, examples, selected_labels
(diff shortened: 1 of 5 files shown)


src/autolabel/tasks/classification.py
```diff
@@ -55,6 +60,40 @@ def construct_prompt(self, input: Dict, examples: List) -> str:
...
+                input_variables=["input"],
+                template="{input}",
+            )
+            label_examples = [{"input": label} for label in labels_list]
+
+            example_selector = SemanticSimilarityExampleSelector.from_examples(
+                # This is the list of labels available to select from.
+                label_examples,
+                # This is the embedding class used to produce embeddings which are used to measure semantic similarity.
+                OpenAIEmbeddings(),
```
rajasbansal, line 76: can we use the embedding model that is the same as the one used for the seed examples, this can be read from the config

src/autolabel/tasks/classification.py
```diff
@@ -55,6 +60,40 @@ def construct_prompt(self, input: Dict, examples: List) -> str:
...
+                # This is the VectorStore class that is used to store the embeddings and do a similarity search over.
+                VectorStoreWrapper(cache=False),
+                # This is the number of examples to produce.
+                k=10,
```
rishabh-bhargava, line 80: We should have this value of `k` be configurable. Maybe 10 is a reasonable default, but we might want workflows in the future that automatically test for the right value of `k`.

src/autolabel/tasks/classification.py
```diff
@@ -55,6 +60,40 @@ def construct_prompt(self, input: Dict, examples: List) -> str:
...
+            )
+            similar_prompt = FewShotPromptTemplate(
+                example_selector=example_selector,
+                example_prompt=example_prompt,
+                prefix="Input: {example}",
```
rajasbansal, line 85: use the example template from the config here? see how the seed examples are prepared

src/autolabel/tasks/classification.py
```diff
@@ -55,6 +60,40 @@ def construct_prompt(self, input: Dict, examples: List) -> str:
...
+
+        # if large number of labels, filter labels_list by similarity of labels to input
+        if num_labels >= 50:
+            example_prompt = PromptTemplate(
+                input_variables=["input"],
+                template="{input}",
+            )
+            label_examples = [{"input": label} for label in labels_list]
+
+            example_selector = SemanticSimilarityExampleSelector.from_examples(
```
rajasbansal, line 72: instead of creating this each time, is it possible to construct this example selector once and then just call sample labels each time. This would make sure that we just embed the label list once.

src/autolabel/tasks/classification.py
```diff
@@ -55,6 +60,40 @@ def construct_prompt(self, input: Dict, examples: List) -> str:
         # prepare task guideline
         labels_list = self.config.labels_list()
         num_labels = len(labels_list)
+
+        # if large number of labels, filter labels_list by similarity of labels to input
+        if num_labels >= 50:
+            example_prompt = PromptTemplate(
+                input_variables=["input"],
```
rajasbansal, line 67: nit: call this label?

src/autolabel/tasks/classification.py
```diff
@@ -55,6 +60,40 @@ def construct_prompt(self, input: Dict, examples: List) -> str:
...
+            )
+            label_examples = [{"input": label} for label in labels_list]
+
+            example_selector = SemanticSimilarityExampleSelector.from_examples(
+                # This is the list of labels available to select from.
+                label_examples,
+                # This is the embedding class used to produce embeddings which are used to measure semantic similarity.
+                OpenAIEmbeddings(),
+                # This is the VectorStore class that is used to store the embeddings and do a similarity search over.
+                VectorStoreWrapper(cache=False),
```
rajasbansal, line 78: went through the code, ideally we don't need to set this cache as False and can use the cache setting from teh config, but if not, this would still be fine.

src/autolabel/tasks/classification.py
```diff
@@ -55,6 +60,40 @@ def construct_prompt(self, input: Dict, examples: List) -> str:
         # prepare task guideline
         labels_list = self.config.labels_list()
         num_labels = len(labels_list)
+
+        # if large number of labels, filter labels_list by similarity of labels to input
+        if num_labels >= 50:
```
rajasbansal, line 65: can we do this based on some config setting. just want to make sure that we have the ability to turn this on or off.  we can do this on num_labels > 50 if the config setting corresponding to this is not set

src/autolabel/tasks/classification.py
rajasbansal, line 80: lets make this configurable

rajasbansal (commented):
did we benchmark this on banking/ledgar. I hope there isn't a big drop in performance using this approach

iomap:
> did we benchmark this on banking/ledgar. I hope there isn't a big drop in performance using this approach

My initial test on Ledgar:
    Previous accuracy: 71% (from Jupyter notebook in autolabel/examples/ledgar)
    Accuracy with change (on 100 samples): 68%

Roughly the same, but need to continue testing. Will try out a full run on banking and ledgar datasets.

src/autolabel/tasks/classification.py
```diff
@@ -55,6 +60,40 @@ def construct_prompt(self, input: Dict, examples: List) -> str:
...
+                prefix="Input: {example}",
+                suffix="",
+                input_variables=["example"],
+            )
+            sampled_labels = similar_prompt.format(example=input["example"])
+            split_lines = sampled_labels.split("\n")
+            labels_list = []
+            for i in range(1, len(split_lines)):
+                if split_lines[i]:
+                    labels_list.append(split_lines[i])
```
iomap, line 94: This might break for input examples that contain newline characters `\n`.

Maybe I do a check that `split_lines[i] in labels_list` before appending

src/autolabel/tasks/classification.py
```diff
@@ -55,6 +60,40 @@ def construct_prompt(self, input: Dict, examples: List) -> str:
...
+                # This is the number of examples to produce.
+                k=10,
+            )
+            similar_prompt = FewShotPromptTemplate(
+                example_selector=example_selector,
+                example_prompt=example_prompt,
+                prefix="Input: {example}",
+                suffix="",
+                input_variables=["example"],
+            )
```
nihit, line 88: please revamp this implementation. 

1. No need to go via FewShotExampleTemplate and semantic similarity example selector. 
2. The label selection should conceptually consist of 3 steps: (i) input row --> formatted example (ii) compute embedding of the formatted example (iii) find nearest neighbors from among the label list
3. The embeddings for labels in the label list should be computed just once, not once per row

src/autolabel/tasks/classification.py
nihit: +1

src/autolabel/tasks/classification.py
nihit: yes, we should enable this "label selection" from a config parameter, not a hardcoded num_labels threshold

src/autolabel/tasks/classification.py
nihit: This should not be needed once the implementation is revamped

nihit (changes requested):
I suggest abstracting all this logic away in a "Label Selector" class, similar to what the Example Selectors do. 

The Label Selector would be initialized once in the run/plan agent method by reading appropriate fields from the config (again, very similar to example selector)

The agent can then call this object's "select labels" function when labeling each example to get a list of K most likely labels, like https://github.com/refuel-ai/autolabel/blob/main/src/autolabel/labeler.py#L183

And pass it to the task object (like https://github.com/refuel-ai/autolabel/blob/main/src/autolabel/labeler.py#L189)

src/autolabel/tasks/classification.py
iomap: done in [f97a82d](https://github.com/refuel-ai/autolabel/pull/561/commits/f97a82d57b03bc3da5c59604b23ed7a573d230f7)

can now specify value of k in autolabel config file like so:
```
        "label_selection": true,
        "label_selection_count": 10
```

src/autolabel/tasks/classification.py
iomap: done in latest commit.

It now chooses embedding function based on `config.embedding_provider()`
```
            self.label_selector = LabelSelector.from_examples(
                labels=self.config.labels_list(),
                k=self.config.label_selection_count(),
                embedding_func=PROVIDER_TO_MODEL.get(
                    self.config.embedding_provider(), DEFAULT_EMBEDDING_PROVIDER
                )(),
            )
```

src/autolabel/tasks/classification.py
iomap: This was unnecessary and has been removed in latest commit.

src/autolabel/tasks/classification.py
iomap: Good idea.

I now construct the LabelSelector once in agent.run() and agent.plan(). This way embeddings of labels are only computed once.

src/autolabel/tasks/classification.py
iomap: removed in latest commit

src/autolabel/tasks/classification.py
iomap: done in [f97a82d](https://github.com/refuel-ai/autolabel/pull/561/commits/f97a82d57b03bc3da5c59604b23ed7a573d230f7)

can now turn on/off label selection in config (as well as the number of labels to select), like so:
```
        "label_selection": true,
        "label_selection_count": 10
```

src/autolabel/tasks/classification.py
iomap: done in latest commit.

src/autolabel/tasks/classification.py
iomap: correct, this has been removed.

src/autolabel/tasks/classification.py
iomap: I have revamped the implementation.

The FewShotExampleTemplate and semantic similarity example selector has been removed entirely.

Embeddings for labels in the label list is computed only once, in agent.plan() and agent.run()

iomap:
Worth noting that after refactoring this PR, I am noticing a slight (~5%) impact on labeling accuracy on Ledgar dataset

Results with label_selection = true

┏━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━━━━━┓
┃ accuracy ┃ support  ┃ completion_rate     ┃
┡━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━━━┩
.  │ 0.6465       │ 100             │ 0.99                            │
└──────────┴─────────┴─────────────────┘

Results with label_selection = false

┏━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━━━━━┓
┃ accuracy ┃ support ┃ completion_rate ┃
┡━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━━━┩
│ 0.7071   │ 100     │ 0.99            │
└──────────┴─────────┴─────────────────┘

Prior to the revamp, accuracy was roughly equal in both cases.

Perhaps our cos_sim() function isn't quite as good as the langchain similarity selector I was using previously? That or the embedding function is configured differently.

iomap:
How I tested:

```
import os
# provide your own OpenAI API key here
os.environ['OPENAI_API_KEY'] = '...'

from autolabel import get_data

get_data('ledgar')

import json

from autolabel import LabelingAgent

# load the config
with open('config_ledgar.json', 'r') as f:
     config = json.load(f)

# create an agent for labeling
agent = LabelingAgent(config=config)

from autolabel import AutolabelDataset
ds = AutolabelDataset("test.csv", config=config)
agent.plan(ds)

# now, do the actual labeling
ds = agent.run(ds, output_name="output_test.csv", max_items=100)
```

`config_ledgar.json`
```
{
    "task_name": "LegalProvisionsClassification",
    "task_type": "classification",
    "dataset": {
        "label_column": "label",
        "delimiter": ","
    },
    "model": {
        "provider": "openai",
        "name": "gpt-3.5-turbo"
    },
    "prompt": {
        "task_guidelines": "You are an expert at understanding legal contracts. Your job is to correctly classify legal provisions in contracts into one of the following categories.\nCategories:{labels}\n",
        "labels": [
            "Agreements",
            "Amendments",
            "Adjustments",
            "Anti-Corruption Laws",
            "Applicable Laws",
            "Approvals",
            "Arbitration",
            "Assignments",
            "Assigns",
            "Authority",
            "Authorizations",
            "Base Salary",
            "Benefits",
            "Binding Effects",
            "Books",
            "Brokers",
            "Capitalization",
            "Change In Control",
            "Closings",
            "Compliance With Laws",
            "Confidentiality",
            "Consent To Jurisdiction",
            "Consents",
            "Construction",
            "Cooperation",
            "Costs",
            "Counterparts",
            "Death",
            "Defined Terms",
            "Definitions",
            "Disability",
            "Disclosures",
            "Duties",
            "Effective Dates",
            "Effectiveness",
            "Employment",
            "Enforceability",
            "Enforcements",
            "Entire Agreements",
            "Erisa",
            "Existence",
            "Expenses",
            "Fees",
            "Financial Statements",
            "Forfeitures",
            "Further Assurances",
            "General",
            "Governing Laws",
            "Headings",
            "Indemnifications",
            "Indemnity",
            "Insurances",
            "Integration",
            "Intellectual Property",
            "Interests",
            "Interpretations",
            "Jurisdictions",
            "Liens",
            "Litigations",
            "Miscellaneous",
            "Modifications",
            "No Conflicts",
            "No Defaults",
            "No Waivers",
            "Non-Disparagement",
            "Notices",
            "Organizations",
            "Participations",
            "Payments",
            "Positions",
            "Powers",
            "Publicity",
            "Qualifications",
            "Records",
            "Releases",
            "Remedies",
            "Representations",
            "Sales",
            "Sanctions",
            "Severability",
            "Solvency",
            "Specific Performance",
            "Submission To Jurisdiction",
            "Subsidiaries",
            "Successors",
            "Survival",
            "Tax Withholdings",
            "Taxes",
            "Terminations",
            "Terms",
            "Titles",
            "Transactions With Affiliates",
            "Use Of Proceeds",
            "Vacations",
            "Venues",
            "Vesting",
            "Waiver Of Jury Trials",
            "Waivers",
            "Warranties",
            "Withholdings"
        ],
        "example_template": "Example: {example}\nOutput: {label}",
        "few_shot_examples": "seed.csv",
        "few_shot_selection": "semantic_similarity",
        "few_shot_num": 4,
        "label_selection": false,
        "label_selection_count": 10
    }
}
```

src/autolabel/tasks/classification.py
```diff
@@ -21,6 +21,11 @@
 
 import json
 
+from langchain.prompts.example_selector import SemanticSimilarityExampleSelector
```
rajasbansal, line 24: probably don't need these imports now?

rajasbansal (approved):
lgtm!

src/autolabel/labeler.py
```diff
@@ -346,7 +376,13 @@ def plan(
                 )
             else:
                 examples = []
-            final_prompt = self.task.construct_prompt(input_i, examples)
+            if self.config.label_selection():
```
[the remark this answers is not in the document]
iomap: good catch. Will check that it is a classification task (if label_selection = true)

src/autolabel/tasks/classification.py
iomap: removed these imports in [e5ff12a](https://github.com/refuel-ai/autolabel/pull/561/commits/e5ff12a1a3a674de746ad9f49b55511486c15f9d)

src/autolabel/labeler.py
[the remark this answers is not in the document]
iomap: done in [2a6ec29](https://github.com/refuel-ai/autolabel/pull/561/commits/2a6ec296cb75347fc662bc0bf248489e00bc1e08)

nihit (approved):

nihit:
@iomap to followup with any updates to documentation

Issue #473, opened 2023-07-20
[Bug]: Give the user the ability to disable quoting when reading csv


Tyrest:
**Describe the bug**
Some csv files will have rows that only have one `"` character, causing issues with reading the dataset. `pandas` has the option to treat quote characters as regular characters but autolabel does not expose this option to the user in any way.

**To Reproduce**
Create a new csv with a single quote starting in a row and another single quote ending in a proceeding row. For example:
| example | label |
| ------- | ----- |
| sure | "yes |
| maybe | no" |
| ok | yes |

Without the quoting option, autolabel should read this csv as

| example | label |
| ------- | ----- |
| sure | "yes\nmaybe,no" |
| ok | yes |

**Expected behavior**
We should expose this option to the user (probably in the dataset section of the config file) and be able to read csv files that have this issue.

PR #563, opened 2023-09-05
Support disable quoting while reading .csv files

mkchaitanya03:
Addresses Issue #473 
Optionally disable quoting while reading .csv files.


nihit (approved):
looks good! thanks for the contribution

PR #570, opened 2023-09-10
Add support for Redis Cache

yadavsahil197:
Adding support for Redis as an cache endpoint. The Redis cache endpoint can be used for both generations and transformations by passing an instance of the RedisCache to the Labeling Agent.


pyproject.toml
```diff
@@ -28,6 +28,7 @@ dependencies = [
     "scikit-learn >= 1.0.0",
     "tenacity >= 8.2.2",
     "SQLAlchemy >= 2.0.19",
+    "redis >= 3.5.3",
```
nihit, line 31: let's make this an optional (extension) dependency ?

nihit (approved):
lgtm otherwise

src/autolabel/labeler.py
```diff
@@ -53,15 +57,28 @@ class LabelingAgent:
...
-        self.generation_cache = SQLAlchemyGenerationCache() if cache else None
-        self.transform_cache = SQLAlchemyTransformCache() if cache else None
+        self.generation_cache = generation_cache
+        self.transform_cache = transform_cache
+        if cache:
+            logger.warning(
+                f"cache parameter is deprecated and will be removed soon. Please use generation_cache and transform_cache instead."
+            )
+            self.generation_cache = (
+                generation_cache if generation_cache else SQLAlchemyGenerationCache()
```
rajasbansal, line 76: What would be the way to disable caches if cache parameter is going to be depracated?

src/autolabel/schema.py
```diff
@@ -209,6 +210,24 @@ class GenerationCacheEntry(BaseModel):
...
+        """
+        Generates a unique ID for the given generation cache configuration
+        """
+        return calculate_md5([self.model_name, self.model_params, self.prompt])
+
+    def get_serialized_output(self) -> str:
+        """
+        Returns the serialized cache entry output
+        """
+        return json.dumps([gen.dict() for gen in self.generations])
```
rajasbansal, line 223: For things like bytes objects which come as a result of transforms, json dumping fails. That is why we switched to serializing using pickle.

src/autolabel/labeler.py
yadavsahil197: The user can pass the cache objects directly to generation_cache and transform_cache to use the cache if needed.

src/autolabel/schema.py
yadavsahil197: Makes sense, I can move to pickle in a follow up PR

PR #577, opened 2023-09-16
Cache confidence scores

mkchaitanya03:
Supports caching confidence scores.

### DIFF
--- a/src/autolabel/cache/sqlalchemy_confidence_cache.py
+++ b/src/autolabel/cache/sqlalchemy_confidence_cache.py
@@ -33,10 +33,10 @@ class SQLAlchemyConfidenceCache(BaseCache):
         cache_entry = ConfidenceCacheEntryModel.get(self.session, entry)
         if cache_entry is None:
             logger.debug("Cache miss")
-            return []
+            return None
 
         logger.debug("Cache hit")
-        return cache_entry.confidence_score
+        return cache_entry.logprobs
 
     def update(self, entry: ConfidenceCacheEntry) -> None:
         """Inserts the provided ConfidenceCacheEntry into the Cache, overriding it if it already exists
--- a/src/autolabel/cache/sqlalchemy_confidence_cache.py
+++ b/src/autolabel/cache/sqlalchemy_confidence_cache.py
@@ -1,0 +1,50 @@
+from sqlalchemy.orm import sessionmaker
+from typing import Optional
+from autolabel.schema import ConfidenceCacheEntry
+from autolabel.database import create_db_engine
+from autolabel.data_models import Base
+from .base import BaseCache
+from autolabel.data_models import ConfidenceCacheEntryModel
+import logging
+
+logger = logging.getLogger(__name__)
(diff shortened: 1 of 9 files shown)


src/autolabel/cache/sqlalchemy_confidence_cache.py
```diff
@@ -0,0 +1,50 @@
...
+        """Retrieves an entry from the Cache. Returns None if not found.
+        Args:
+            entry: ConfidenceCacheEntry we wish to retrieve from the Cache
+        Returns:
+            result: A floating point number representing the confidence score for this generation. None if not found.
+        """
+        cache_entry = ConfidenceCacheEntryModel.get(self.session, entry)
+        if cache_entry is None:
+            logger.debug("Cache miss")
+            return []
```
rajasbansal, line 36: change default return type, should be a float or None

src/autolabel/confidence.py
```diff
@@ -188,6 +203,33 @@ def compute_confidence(self, model_input, model_output) -> Union[dict, List[dict
             )
             return [{"": 0.5}]
 
+    def get_cached_confidence(self, model_generation: LLMAnnotation) -> Optional[float]:
+        cache_entry = ConfidenceCacheEntry(
+            prompt=model_generation.prompt,
+            raw_response=model_generation.raw_response,
+            score_type=self.score_type,
+        )
+        confidence = self.cache.lookup(cache_entry)
```
rajasbansal, line 212: what if there is a cache miss?

src/autolabel/confidence.py
```diff
@@ -188,6 +203,33 @@ def compute_confidence(self, model_input, model_output) -> Union[dict, List[dict
...
+        if confidence:
+            if self.score_type == "logprob_average" or "p_true":
+                confidence = confidence[model_generation.raw_response]
```
rajasbansal, line 215: Why do we need to go into the dictionary?

src/autolabel/confidence.py
```diff
@@ -149,6 +161,9 @@ def calculate(self, model_generation: LLMAnnotation, **kwargs) -> float:
             keys=keys,
             **kwargs,
         )
+        model_generation.confidence_score = confidence
+        if self.cache:
+            self.update_cache(model_generation)
```
rajasbansal, line 166: this should only be called in case of a cache miss

src/autolabel/confidence.py
```diff
@@ -149,6 +168,18 @@ def calculate(self, model_generation: LLMAnnotation, **kwargs) -> float:
             keys=keys,
             **kwargs,
         )
+        model_generation.confidence_score = confidence
+
+        # On cache miss, update cache
```
rajasbansal, line 173: since caching is only done for the case where the llm does not return logprobs - lets move the code in that if condition. here in the case we have the llm return logprobs then we will have cache_miss variable not defined error

rajasbansal (approved):
lgtm otherwise

PR #584, opened 2023-09-29
Expanding benchmark

DhruvaBansal00:
This PR adds the following 4 datasets as examples for using autolabel:

1. Craigslist bargains: Given a passage consisting of negotiation between a seller and buyer about sale of an item, classify the item into one of 'housing', 'furniture', 'bike', 'phone', 'car', 'electronics'.

2. Mathdatasets: Given a math question, classify it into algebra, arithmetic, measurement, numbers, and probability. The guidelines for each class are also provided in the instructions.

3. Lexical relation cogalexv: Given two words, classify the relationship between them into one of ANT, HYPER, PART_OF, SYN, RANDOM

4. Read-extract: Given a paragraph, extract the exact substring from the context which answers the question in the prompt.

### DIFF
--- a/examples/banking/example_banking.ipynb
+++ b/examples/banking/example_banking.ipynb
@@ -981,7 +981,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.8.9"
+   "version": "3.10.6"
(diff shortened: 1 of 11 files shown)


nihit (approved):
🚀

PR #585, opened 2023-10-02
Allow additional properties

Abhinav-Naikawadi:


nihit (approved):

Abhinav-Naikawadi:
Instead of merging this change, callers can set validate to False here: https://github.com/refuel-ai/autolabel/blob/32c55e2b438d635e6aae2efb8df51e1b152ddef7/src/autolabel/configs/config.py#L60C64-L60C64

PR #586, opened 2023-10-02
add documentation for using label_selection on large classification tasks

iomap:


iomap:
What documentation looks like with these changes.
image

nihit (approved):

PR #605, opened 2023-10-18
Input columns in autolabel config

Abhinav-Naikawadi:


nihit:
@Abhinav-Naikawadi we also need to add it here? https://github.com/refuel-ai/autolabel/blob/main/src/autolabel/configs/schema.py

rajasbansal (commented):
@Abhinav-Naikawadi use this in few shot as well?

rajasbansal (approved):

PR #614, opened 2023-10-26
Filter labels not in labels list for multilabel classification

Abhinav-Naikawadi:


Abhinav-Naikawadi (commented):
Filter out labels that are not in labels list and only output NO_LABEL when there are no labels in labels list

nihit (approved):

PR #625, opened 2023-11-10
Multimodal Labeling

DhruvaBansal00:
Adding initial support for sending in images alongside text to GPT-4V


DhruvaBansal00:
This PR only works for classification at the moment. Don't think it's in a state where we can merge it but getting it up so we can discuss changes during the demo.

PR #629, opened 2023-11-15
Support few shot examples with attribute extraction tasks

Abhinav-Naikawadi:


rishabh-bhargava:
@Abhinav-Naikawadi is this ready for review? Cc: @rajasbansal 

I will be doing an extraction demo tomorrow morning (no PDFs) and would be good to have this ready by then! 

Happy to review

nihit:
@rishabh-bhargava it's not just this change, we need a couple of other changes in the platform as well

nihit (approved):

PR #635, opened 2023-11-17
Chunking for confidence if context exceeds context length

rajasbansal:
In this PR, we will chunk confidence prompts in order to support getting confidence from a model having context length smaller than the context length supported by the generation model. We do this by chunking the document into smaller chunks and sending the prompt along with the smaller chunks of the document. we support multiple aggregation functions over the confidence chunks. 

For extraction, max makes more sense than mean which might make more sense for a classification task

### DIFF
--- a/src/autolabel/labeler.py
+++ b/src/autolabel/labeler.py
@@ -301,6 +301,7 @@ class LabelingAgent:
                                 key_to_chunk = None
                                 for key in chunk.keys():
                                     # TODO(rajas): Better way to find the key to chunk
+                                    # Potentially take this as an input from the user
                                     if (
                                         self.get_num_tokens(chunk[key])
                                         > num_tokens_per_chunk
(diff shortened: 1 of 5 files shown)


src/autolabel/labeler.py
```diff
@@ -651,3 +734,13 @@ def get_all_annotations(self):
             return [pickle.loads(a.llm_annotation) for a in db_result]
         else:
             return self.all_annotations
+
+    def get_num_tokens(self, inp: str) -> int:
+        """Returns the number of tokens in the prompt"""
+        return len(self.confidence_tokenizer.encode(str(inp)))
+
+    def chunk_string(self, inp: str, chunk_size: int) -> List[str]:
```
nihit, line 742: move to utils.py ?

nihit (approved):
lgtm

src/autolabel/labeler.py
rajasbansal: uses the tokenizer, for encoding and decoding so kept it here

PR #643, opened 2023-11-22
Supporting confidence calculation using RefuelLLM

DhruvaBansal00:
Screenshot 2023-11-22 at 1 55 11 PM

Confidence calculation will now call https://llm.refuel.ai/models/refuelllm/confidence

### DIFF
--- a/src/autolabel/tasks/base.py
+++ b/src/autolabel/tasks/base.py
@@ -88,12 +88,19 @@ class BaseTask(ABC):
         return self.config.few_shot_algorithm() in [x.value for x in FewShotAlgorithm]
 
     @abstractmethod
-    def construct_prompt(self, input: str, examples: List) -> str:
+    def construct_prompt(
+        self,
+        input: str,
+        examples: List,
+        prompt_template_override: PromptTemplate = None,
+        refuel_prompt_override: bool = False,
+        output_guidelines_override: str = None,
+        **kwargs,
+    ) -> str:
         pass
 
-    def construct_refuel_prompt(self, input: str, examples: List) -> str:
-        self.use_refuel_prompt_schema = True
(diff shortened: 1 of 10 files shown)


DhruvaBansal00:
Quick summary of changes: we have a new method called `construct_refuel_prompt` in base task which will set the right task properties and call `construct_prompt` to get the refuel llm prompt. We call this right before making a call to the confidence api, add it to the LLMAnnotation object and then compute confidence using this prompt.

DhruvaBansal00:
Screenshot 2023-11-22 at 2 13 30 PM

Working for QA as well now

src/autolabel/tasks/base.py
```diff
@@ -94,6 +91,22 @@ def _is_few_shot_mode(self) -> bool:
...
+        self.use_refuel_prompt_schema = True
+        self.prompt_template = PromptTemplate(
+            input_variables=get_format_variables(
+                self.FEW_SHOT_TEMPLATE_REFUEL_LLM
+                if self._is_few_shot_mode()
+                else self.ZERO_SHOT_TEMPLATE_REFUEL_LLM
+            ),
+            template=self.example_template,
+        )
+        refuel_prompt = self.construct_prompt(input, examples)
```
nihit, line 104: hmm, i suggest instead passing this flag as a parameter to ` self.construct_prompt(...)` to denote which "prompt modes" to construct - the default one or the one customized for refuel llm. that seems cleaner than setting some isntance variavle, and then resetting it again.

nihit (approved):
lgtm otherwise

DhruvaBansal00:
@nihit re-requesting review based on our discussion. Have tested this implementation with qa and classification tasks as well.

nihit (approved):

PR #645, opened 2023-11-22
Add metadata like cost, latency, token count to llm annotations

yadavsahil197:


src/autolabel/models/palm.py
```diff
@@ -120,17 +135,21 @@ def _label(self, prompts: List[str]) -> RefuelLLMResult:
...
                         self.SEP_REPLACEMENT_TOKEN, "\n"
                     )
             return RefuelLLMResult(
-                generations=result.generations, errors=[None] * len(result.generations)
+                generations=result.generations,
+                errors=[None] * len(result.generations),
+                latencies=[end_time - start_time] * len(result.generations),
             )
         except Exception as e:
-            self._label_individually(prompts)
```
rajasbansal, line 133: this might no longer be needed given we are doing the lableing with chunk size as 1 - independent of this PR

src/autolabel/models/palm.py
```diff
@@ -120,17 +135,21 @@ def _label(self, prompts: List[str]) -> RefuelLLMResult:
             prompts = [[HumanMessage(content=prompt)] for prompt in prompts]
 
         try:
+            start_time = time()
             result = self._label_with_retry(prompts)
+            end_time = time()
```
rajasbansal, line 140: want to make sure we want to record time including retries

rajasbansal (approved):

rajasbansal:
lgtm after tests pass

src/autolabel/models/palm.py
yadavsahil197: Good catch, We want to keep the latency independent of the retry, so should only include the time for a successful call.

PR #661, opened 2023-12-13
parse attribute extraction confidence based on finding tokens

rajasbansal:
Find the tokens in the confidence returned logprobs and use these to make confidence finding more robust

### DIFF
--- a/src/autolabel/confidence.py
+++ b/src/autolabel/confidence.py
@@ -83,24 +83,42 @@ class ConfidenceCalculator:
         This function calculates the confidence score per key. This will return
         a confidence score per key.
         """
-
-        # Find all '"' and '",'
-        # This is a hacky way to find all the keys in the prompt
-        indices = []
-        for ind, logprob in enumerate(logprobs):
-            key = list(logprob.keys())[0]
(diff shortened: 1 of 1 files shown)


src/autolabel/confidence.py
```diff
@@ -83,24 +83,42 @@ def logprob_average_per_key(
...
-            logger.error("Unable to find all keys in prompt")
-            return {key: 0 for key in keys}
-
         # Find the logprob for each key
         logprob_per_key = {}
-        for i, key in enumerate(keys):
-            logprob_per_key[key] = self.logprob_average(
-                logprobs[indices[4 * i + 2] + 1 : indices[4 * i + 3]]
+
+        mapping = []
```
nihit, line 89: can you share some more details about this approach ? Will improve readability a lot in the future - 
1. Start with an example input for logprobs and keys list
2. For this, what will `mapping` and `full_string` values be?
3. What will `locations` be ?

rajasbansal: added more details based on a confidence example string. Lmk if this is helpful

nihit (approved):

PR #670, opened 2023-12-22
Allowing explanation generation for attribute extraction

DhruvaBansal00:
Adding explanation support for attribute extraction

### DIFF
--- a/src/autolabel/configs/config.py
+++ b/src/autolabel/configs/config.py
@@ -135,7 +135,7 @@ class AutolabelConfig(BaseConfig):
 
     def explanation_column(self) -> str:
         """Returns the name of the column containing an explanation as to why the data is labeled a certain way"""
-        return self._dataset_config.get(self.EXPLANATION_COLUMN_KEY, None)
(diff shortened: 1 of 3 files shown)


src/autolabel/configs/config.py
```diff
@@ -135,7 +135,7 @@ def input_columns(self) -> List[str]:
 
     def explanation_column(self) -> str:
         """Returns the name of the column containing an explanation as to why the data is labeled a certain way"""
-        return self._dataset_config.get(self.EXPLANATION_COLUMN_KEY, None)
+        return self._dataset_config.get(self.EXPLANATION_COLUMN_KEY, "explanation")
```
rajasbansal, line 138: constant

rajasbansal (approved):
lgtm

src/autolabel/configs/config.py
DhruvaBansal00: Will make a constant. Thanks!

DhruvaBansal00: Actually, do we have a separate file for storing constants for default config values? I think this might be it.

PR #700, opened 2024-02-02
Minor fixes to the doc and user experience elements

vaibagra:
# Pull Review Summary

## Description

There was some data inconsistency in the way that the movie_reviews dataset was set up so I changed the s3 bucket to contain the updated test.csv (with the correct labels). Additionally added some error handling for non existent keys. Finally changed get_data() slightly to create folders to store the downloaded data in folders data/{dataset_name} to allow multiple datasets to be operated simultaneously. Updated the docs on the relevant pages to reflect this change. 

## Type of change

- Cleaned some code
- Added some errors for non existent keys for example_template

## Tests

Locally ran and tested different non existent keys to ensure that the correct error is thrown.

### DIFF
--- a/src/autolabel/utils.py
+++ b/src/autolabel/utils.py
@@ -335,12 +335,7 @@ def get_data(dataset_name: str, force: bool = False):
 
     def download(url: str) -> None:
         """Downloads the data given an url"""
-        os.mkdir("data") if not os.path.exists("data") else None
-        (
-            os.mkdir(f"data/{dataset_name}")
-            if not os.path.exists(f"data/{dataset_name}")
-            else None
-        )
+        os.makedirs(os.path.join("data", dataset_name), exist_ok=True)
         file_name = os.path.join("data", dataset_name, os.path.basename(url))
         if force and os.path.exists(file_name):
             print(f"File {file_name} exists. Removing")


src/autolabel/tasks/attribute_extraction.py
```diff
@@ -120,8 +120,19 @@ def construct_prompt(
 
         input[self.OUTPUT_DICT_KEY] = ""
 
+        # check if all mapped keys in input are in the example template
+        try:
+            current_example = example_template.format(**input)
+        except KeyError as e:
+            curr_text_prompt = example_template.format_map(defaultdict(str, input))
```
rajasbansal, line 127: we should be updating current_example as well right? I think it being used in the trim_prompt function below

tests/unit/tasks_test.py
```diff
@@ -84,23 +84,19 @@ def test_classification_no_label_column_in_input():
 def test_classification_no_label_column_in_config():
```
rajasbansal, line 84: I think this test itself can be deleted. it doesn't make sense to set it to nothing

tests/unit/test_utils.py
```diff
@@ -1,36 +1,42 @@
...
+from autolabel import utils
+
 console = Console()
 
 
 def test_get_data(mocker) -> None:
     """Test Get Data"""
     dataset_name = "banking"
 
+    os.mkdir("data") if not os.path.exists("data") else None
```
rajasbansal, line 18: i think there is an argument in mkdir to handle this  called exist_ok

rajasbansal (commented):
lgtm - need some changes to examples

rajasbansal (approved):
lgtm

PR #713, opened 2024-02-13
Added better functionality for label selection

### LINKED ISSUE #712: Handle larger taxonomies with top k label selection to pass to the LLM

Vaibhav2001:
# Pull Review Summary
Continued from https://github.com/refuel-ai/autolabel/pull/561. This closes #712

## Description

- Now we can select the top K labels to pass to the LLM with `label_selection`, and specify the count though `label_selection_count`, and a threshold via `label_selection_threshold`. 
- Enabled caching for storing embeddings. 
- Few shot examples will only contain the correct examples which have the provided label.  (only works with `"fixed"` algorithm for now)

**Benefits**: 

- Leads to a similar accuracy as base classification, however leads to a better completion rate when num_labels is huge, as the model prediction does not map exactly with any label (mostly because of large context length).
- Cost reduction as the number of tokens passed to the final model can be small. 

**Usage:**

```
config = {
    "task_name": "BankingComplaintsClassification",
    "task_type": "classification",
        ...
        "labels": label_descriptions,
        "label_selection": True,
        "label_selection_count": 10,
        "label_selection_threshold": 0.5,
        ...
    },
}
```

- Can use label descriptions to embed the labels. 
- `"label_selection_count": -1` makes maximum possible labels equal to `len(labels)`. Defaults to `min(10, num_lables)`
- `label_selection_threshold` enables us to only choose the labels where `label_score / topScore > threshold` for effective filtering. This value defaults to 0.95 (based on experimentation). 

## Type of change

- Improving existing feature 

## Tests

Ran benchmarking tests for different datasets. 

## Future Work 

- Do not classify if there is only one selected_label. Needs to change the annotation processing in `dataset.process_labels. `
- Costing for embedding model. However this is typically upper bounded by the base model and is cheaper by a factor of up to 1000.


src/autolabel/labeler.py
```diff
@@ -221,6 +222,14 @@ def run(
             )
 
         if self.example_selector is None:
+            if (
+                self.config.label_selection()
+                and self.config.few_shot_algorithm() != "fixed"
```
rajasbansal, line 227: semantic similarity will be very important to support here

src/autolabel/labeler.py
```diff
@@ -234,54 +243,63 @@ def run(
...
-                final_prompt = self.task.construct_prompt(
-                    chunk,
-                    examples,
-                    selected_labels=selected_labels,
-                    max_input_tokens=self.llm.DEFAULT_CONTEXT_LENGTH,
-                    get_num_tokens=self.llm.get_num_tokens,
-                )
+                # get every column except the one we want to label
+                toEmbed = chunk.copy()
+                del toEmbed[self.config.label_column()]
```
rajasbansal, line 274: what if the label column doesnt exist in the incoming chunk, i.e we are dealing with an unlabeled dataset?

src/autolabel/labeler.py
```diff
@@ -234,54 +243,63 @@ def run(
...
+                if self.example_selector:
+                    examples = self.example_selector.select_examples(
+                        safe_serialize_to_string(chunk),
+                    )
+
+            # Construct Prompt to pass to LLM
+            final_prompt = self.task.construct_prompt(
+                chunk,
+                examples,
+                selected_labels=selected_labels,
```
rajasbansal, line 299: does construct prompt support selected_labels as an argument already?

src/autolabel/few_shot/label_selector.py
```diff
@@ -13,45 +17,56 @@ class LabelSelector:
...
     def select_labels(self, input: str) -> List[str]:
         """Select which labels to use based on the similarity to input"""
-        input_embedding = self.embedding_func.embed_query(input)
+        input_embedding = self.vectorStore._get_embeddings([input])
 
         scores = []
         for label, embedding in self.labels_embeddings.items():
             similarity = cos_sim(embedding, input_embedding)
             # insert into scores, while maintaining sorted order
             bisect.insort(scores, (similarity, label))
```
rajasbansal, line 65: lets do this in a vectorized way? not sure how much of a performance hit doing this is

src/autolabel/configs/config.py
```diff
@@ -238,7 +239,16 @@ def label_selection(self) -> bool:
...
+        k = self._prompt_config.get(self.LABEL_SELECTION_COUNT_KEY, 10)
+        if k < 1:
+            return len(self.labels_list())
+        return k
+
+    def label_selection_threshold(self) -> float:
+        """Returns the threshold for label selection in LabelSelector
+        If the similarity score ratio with the top Score is above this threshold,
+        the label is selected."""
+        return self._prompt_config.get(self.LABEL_SELECTION_THRESHOLD, 0.95)
```
rajasbansal, line 251: lets set this to be 0 by default? We should use the threshold only if provided.

src/autolabel/configs/config.py
```diff
@@ -238,7 +239,16 @@ def label_selection(self) -> bool:
 
     def label_selection_count(self) -> int:
```
rajasbansal, line 240: should we call it max selected labels

src/autolabel/labeler.py
Vaibhav2001: yeah

src/autolabel/configs/config.py
Vaibhav2001: got it

src/autolabel/few_shot/label_selector.py
```diff
@@ -46,23 +47,22 @@ def __init__(
...
+                self.vectorStore._get_embeddings(descriptions)
+            )
         else:
-            embeddings = self.vectorStore._get_embeddings(self.labels)
-            for i, label in enumerate(labels):
-                self.labels_embeddings[label] = embeddings[i]
+            self.labels_embeddings = torch.Tensor(
+                self.vectorStore._get_embeddings(self.labels)
+            )
+        print(type(self.labels_embeddings))
```
rajasbansal, line 58: remove print

src/autolabel/few_shot/label_selector.py
```diff
@@ -4,6 +4,7 @@
 from collections.abc import Callable
 from typing import Dict, List, Optional, Tuple, Union
 
+import torch
```
rajasbansal, line 7: does base autolabel have pytorch installed? if not we should try to use numpy

src/autolabel/few_shot/label_selector.py
```diff
@@ -46,23 +47,22 @@ def __init__(
...
-            similarity = cos_sim(embedding, input_embedding)
-            # insert into scores, while maintaining sorted order
-            bisect.insort(scores, (similarity, label))
+        input_embedding = torch.Tensor(self.vectorStore._get_embeddings([input]))
+        scores = cos_sim(input_embedding, self.labels_embeddings).view(-1)
+        scores = list(zip(scores, self.labels))
+        scores.sort(key=lambda x: x[0])
 
         # remove labels with similarity score less than self.threshold*topScore
         return [
```
rajasbansal, line 68: I think the list should be filtered first based on the threshold and then we should select scores[-self.k: ]. This is so that we retrieve k items if the number of labels above threshold is > k

src/autolabel/labeler.py
```diff
@@ -271,7 +272,8 @@ def run(
             ):
                 # get every column except the one we want to label
                 toEmbed = chunk.copy()
-                del toEmbed[self.config.label_column()]
+                if self.config.label_column():
```
rajasbansal, line 275: the label column column will be present in the config but it might not be present in the chunk. We should check if self.config.label_column() is in toEmbed

rajasbansal (approved):
lgtm - after small fixes

src/autolabel/few_shot/label_selector.py
Vaibhav2001: we do

PR #720, opened 2024-02-20
Support json mode for openai models

rajasbansal:
# Pull Review Summary

## Description

Support json mode for openai models.

## Type of change

## Tests

locally

### DIFF
--- a/src/autolabel/models/openai.py
+++ b/src/autolabel/models/openai.py
@@ -47,6 +47,13 @@ class OpenAILLM(BaseModel):
         "gpt-4-1106-preview",
         "gpt-4-0125-preview",
     ]
+    JSON_MODE_MODELS = [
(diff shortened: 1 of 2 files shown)


src/autolabel/models/openai.py
```diff
@@ -47,6 +47,13 @@ class OpenAILLM(BaseModel):
         "gpt-4-1106-preview",
         "gpt-4-0125-preview",
     ]
+    JSON_MODE_MODELS = [
```
Abhinav-Naikawadi, line 50: not supported with just gpt-4?

Abhinav-Naikawadi (approved):

src/autolabel/models/openai.py
rajasbansal: nope

PR #736, opened 2024-02-24
Adding support for async labeling for openai

yadavsahil197:
# Pull Review Summary

## Description
Openai calls are generally quite slow and making those api calls in sequence can take a long time, all the while just waiting for the response. To maximize the throughput, adding support to make async calls to openai to begin with. We will add support for other LLM in another PR.
This PR also makes confidence calls to be async as well by default.

## Type of change


nihit (approved):

nihit:
@yadavsahil197 we might need to modify some tests with this change

yadavsahil197:
Thanks for taking care of the failing tests @nihit!

PR #739, opened 2024-02-26
Gemini support and Langchain upgrade

Vaibhav2001:
# Pull Review Summary

## Description

- Basic support for gemini models. 
- Langchain updated to 0.1.9
- Python updated to 3.9

## Type of change

- This change requires a documentation update for langchain, anthropic, and google section
- Maybe have a blogpost about claude-3 and gemini

## Testing

- [x] Proper testing of classification with few shot
- [x] Other edge cases


.github/workflows/test.yaml
```diff
@@ -15,10 +15,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Install Python3.8
+      - name: Install Python3.9
         uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: 3.9
```
rajasbansal, line 21: @yadavsahil197 @Abhinav-Naikawadi this change will need to go in for all other repos as well. do we expect any problems because of this?

src/autolabel/models/google.py
```diff
@@ -0,0 +1,180 @@
...
+                HarmCategory,
+            )
+        except ImportError:
+            raise ImportError(
+                "tiktoken and langchain_google_genai. Please install it with the following command: pip install 'refuel-autolabel[google]'"
+            )
+
+        self.DEFAULT_PARAMS = {
+            "temperature": 0.0,
+            "topK": 3,
```
rajasbansal, line 54: what does topK do here?

src/autolabel/models/google.py
```diff
@@ -0,0 +1,180 @@
...
+                        error_type=ErrorType.LLM_PROVIDER_ERROR, error_message=str(e)
+                    )
+                )
+                latencies.append(0)
+
+        return RefuelLLMResult(
+            generations=generations, errors=errors, latencies=latencies
+        )
+
+    def _label(self, prompts: List[str]) -> RefuelLLMResult:
```
rajasbansal, line 130: we don't need to do this for google. Separator replacement token was something done specifically for RefuelLLM in the past. Does not need to be replicated here.

Let's follow models/openai.py

src/autolabel/models/google.py
```diff
@@ -0,0 +1,180 @@
...
+
+        self.tiktoken = tiktoken
+
+    @retry(
+        reraise=True,
+        stop=stop_after_attempt(5),
+        wait=wait_exponential(multiplier=1, min=2, max=10),
+        before_sleep=before_sleep_log(logger, logging.WARNING),
+    )
+    def _label_with_retry(self, prompts: List[str]) -> LLMResult:
```
rajasbansal, line 89: this is needed for refuelLLM which does not do retries by default. Langchain does retries by default for openai. Isn't the same done for google as well?

src/autolabel/models/google.py
```diff
@@ -0,0 +1,180 @@
...
+
+class GoogleLLM(BaseModel):
+    SEP_REPLACEMENT_TOKEN = "@@"
+    CHAT_ENGINE_MODELS = ["gemini-pro"]
+
+    DEFAULT_MODEL = "gemini-pro"
+
+    # Reference: https://ai.google.dev/pricing
+    COST_PER_CHARACTER = {
+        "gemini-pro": 0.000125 / 1000,
```
rajasbansal, line 25: let's add other models which might be supported by this api as well. also input and output token prices are different. We have something to handle this in models/openai.py

rajasbansal (approved):

PR #742, opened 2024-02-28
Changing default output to produce only necessary attributes

DhruvaBansal00:
# Pull Review Summary

## Description

Changing attribute extraction prompt to only produce the necessary keys. Enforcing that certain attributes always exist can be enforced by the user by specifying in task level prompt.

## Type of change

- [X] New feature (change which adds functionality)

## Tests

Locally

### DIFF
--- a/src/autolabel/tasks/attribute_extraction.py
+++ b/src/autolabel/tasks/attribute_extraction.py
@@ -33,7 +33,7 @@ logger = logging.getLogger(__name__)
 class AttributeExtractionTask(BaseTask):
     NULL_LABEL = {}
     DEFAULT_TASK_GUIDELINES = "You are an expert at extracting attributes from text. Given a piece of text, extract the required attributes."
-    DEFAULT_OUTPUT_GUIDELINES = "You will return the extracted attributes as a json with the following keys:\n{attribute_json}"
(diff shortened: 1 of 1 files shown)


Abhinav-Naikawadi (approved):

src/autolabel/tasks/attribute_extraction.py
```diff
@@ -33,7 +33,7 @@
 class AttributeExtractionTask(BaseTask):
     NULL_LABEL = {}
     DEFAULT_TASK_GUIDELINES = "You are an expert at extracting attributes from text. Given a piece of text, extract the required attributes."
-    DEFAULT_OUTPUT_GUIDELINES = "You will return the extracted attributes as a json with the following keys:\n{attribute_json}"
+    DEFAULT_OUTPUT_GUIDELINES = "You will return the extracted attributes as a json with the following keys:\n{attribute_json}. Only output labels for the fields that exist in the provided prompt. Do not output anything for attributes that are irrelevant to the given input."
```
nihit, line 36: let's be consistent in instructing the LLM with fields vs attributes and prompt vs input ? 

Something like only include attributes in the JSON that have a valid extracted value for the given input ?

DhruvaBansal00: Makes sense, have pushed a fix for this!

nihit (approved):

PR #745, opened 2024-02-29
Continue web scrape if query column is missing from row

Abhinav-Naikawadi:
# Pull Review Summary

## Description

Avoid blocking web scrape if a query column doesn't exist in a row. The query template will just not format that query column.

## Type of change

- Improvement

## Tests

Tested locally without query column in row

### DIFF
--- a/src/autolabel/transforms/serp_api.py
+++ b/src/autolabel/transforms/serp_api.py
@@ -99,12 +99,6 @@ class SerpApi(BaseTransform):
         return search_result
 
     async def _apply(self, row: Dict[str, Any]) -> Dict[str, Any]:
-        for col in self.query_columns:
-            if col not in row:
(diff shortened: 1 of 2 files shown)


src/autolabel/transforms/serp_api.py
```diff
@@ -99,12 +99,6 @@ async def _get_result(self, query):
         return search_result
 
     async def _apply(self, row: Dict[str, Any]) -> Dict[str, Any]:
-        for col in self.query_columns:
-            if col not in row:
-                raise TransformError(
-                    TransformErrorType.INVALID_INPUT,
-                    f"Missing query column: {col} in row {row}",
-                )
```
nihit, line 107: please log the error at least

nihit (approved):

PR #752, opened 2024-03-05
Add scheme to domains that are missing it

yadavsahil197:
# Pull Review Summary

## Description

For domains that are missing url scheme (http:// or https://), we should add it by default otherwise this results in errors.

## Type of change


Abhinav-Naikawadi (approved):

Abhinav-Naikawadi (commented):
ah I see this was only added for scrapingbee transform, thanks for adding

PR #761, opened 2024-03-11
Add support for multiple image columns in autolabel

Abhinav-Naikawadi:
# Pull Review Summary

## Description

Add support for multiple image columns when using an LLM supporting multimodal input (for now, just GPT-4-Vision).

TODO: For now, this PR does not remove existing single image_column functionality to preserve backwards compatibility, but we should just remove all single image_column support in favor of the new schema before merging.

## Type of change

## Tests

Tested locally with multiple image columns.


src/autolabel/tasks/base.py
```diff
@@ -51,6 +51,7 @@ class BaseTask(ABC):
     def __init__(self, config: AutolabelConfig) -> None:
         self.config = config
         self.image_col = self.config.image_column()
+        self.image_cols = self.config.image_columns()
```
DhruvaBansal00, line 54: Would be best to merge these two together and deduplicate their references in the PR.

DhruvaBansal00 (approved):

PR #766, opened 2024-03-15
Added support for haiku model from Anthropic.

Vaibhav2001:
# Pull Review Summary

## Description

Added support for haiku model from Anthropic. 

## Type of change

### DIFF
--- a/src/autolabel/models/anthropic.py
+++ b/src/autolabel/models/anthropic.py
@@ -10,7 +10,7 @@ from autolabel.schema import RefuelLLMResult
 
 
 class AnthropicLLM(BaseModel):
-    DEFAULT_MODEL = "claude-instant-v1"
+    DEFAULT_MODEL = "claude-3-haiku-20240307"
     DEFAULT_PARAMS = {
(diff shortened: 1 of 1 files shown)


src/autolabel/models/anthropic.py
```diff
@@ -23,13 +23,15 @@ class AnthropicLLM(BaseModel):
         "claude-2.1": (8 / 1_000_000),
         "claude-3-opus-20240229": (15 / 1_000_000),
         "claude-3-sonnet-20240229": (3 / 1_000_000),
+        "claude-3-haiku-20240307": (0.25 / 1_000_000),
```
rajasbansal, line 26: do they also have a non timestamped name which always points to the latest model like there is gpt-4-turbo and gpt-4-1106-preview

rajasbansal (approved):

src/autolabel/models/anthropic.py
Vaibhav2001: I do not think so. I tried testing that but it is not the case. Even langchain uses these.

PR #767, opened 2024-03-19
Change severity level to warning

Abhinav-Naikawadi:
# Pull Review Summary

## Description

We allow web search transforms to go through even if an expected query column is missing (it just doesn't get formatted in the template). So, this should be warning level.

## Type of change

- Nit


Vaibhav2001 (approved):
lgtm!

rajasbansal (approved):

PR #771, opened 2024-03-21
Changes for benchmarking new models

rajasbansal:
# Pull Review Summary

## Description

Changes for benchmarking new models

## Type of change

## Tests

locally


benchmark/benchmark.py
```diff
@@ -0,0 +1,98 @@
...
+    eval_file_name = f"eval_{args.model}_{args.few_shot}_{args.max_items}.json"
+    eval_file_name = eval_file_name.replace("/", "")
+    eval_result = []
+    for dataset in DATASETS:
+        config = json.load(open(f"configs/{dataset}.json", "r"))
+        config["model"]["name"] = args.model
+        config["model"]["provider"] = MODEL_TO_PROVIDER[args.model]
+        config["prompt"]["few_shot_num"] = args.few_shot
+        if not args.few_shot:
+            config["prompt"]["few_shot_selection"] = "fixed"
```
DhruvaBansal00, line 74: Few shot prompting shouldn't be done at all if few shot is 0, correct? Just want to make sure that is the case. We shouldn't have extraneous phrases like "Here are a few demonstrations :<nothing>" if few shot is 0.

benchmark/benchmark.py
```diff
@@ -0,0 +1,98 @@
...
+]
+
+MODEL_TO_PROVIDER = {
+    "gpt-3.5-turbo": "openai",
+    "gpt-4": "openai",
+    "claude-3-opus-20240229": "anthropic",
+    "claude-3-sonnet-20240229": "anthropic",
+    "mistralai/Mistral-7B-v0.1": "vllm",
+    "mistralai/Mistral-7B-Instruct-v0.1": "vllm",
+    "mistralai/Mixtral-8x7B-v0.1": "mistral",
```
DhruvaBansal00, line 38: The provider here should be vllm as well?

src/autolabel/models/vllm.py
```diff
@@ -0,0 +1,93 @@
+import json
+import os
+import requests
+import logging
+from typing import List, Optional, Tuple
+from time import time
+from vllm import LLM, SamplingParams
```
DhruvaBansal00, line 7: We might want t add this to autolabel requirements as well.

src/autolabel/models/vllm.py
```diff
@@ -0,0 +1,93 @@
...
+        self.llm = LLM(model=self.config.model_name())
+        self.model_name = self.config.model_name()
+
+    def _label(self, prompts: List[str]) -> RefuelLLMResult:
+        generations = []
+        errors = []
+        latencies = []
+        for prompt in prompts:
+            try:
+                print(prompt)
```
DhruvaBansal00, line 55: remove?

src/autolabel/models/vllm.py
```diff
@@ -0,0 +1,93 @@
...
+                response = self.llm.generate(prompt, self.params, use_tqdm=False)
+                generations.append(
+                    [Generation(
+                        text=response[0].outputs[0].text.strip(),
+                    )]
+                )
+                print(response[0].outputs[0].text.strip())
```
DhruvaBansal00, line 62: remove?

src/autolabel/models/vllm.py
```diff
@@ -0,0 +1,93 @@
...
+from autolabel.cache import BaseCache
+from autolabel.schema import LabelingError, ErrorType, RefuelLLMResult
+
+from langchain.schema import Generation
+
+UNRETRYABLE_ERROR_CODES = [400, 422]
+logger = logging.getLogger(__name__)
+
+
+class UnretryableError(Exception):
```
DhruvaBansal00, line 24: Are we using this anywhere?

src/autolabel/models/vllm.py
```diff
@@ -0,0 +1,93 @@
...
+        return RefuelLLMResult(generations=generations, errors=errors, latencies = latencies)
+
+    def get_cost(self, prompt: str, label: Optional[str] = "") -> float:
+        return 0
+
+    def returns_token_probs(self) -> bool:
+        return False
+
+    def get_num_tokens(self, prompt: str) -> int:
+        return len(prompt)
```
DhruvaBansal00, line 86: If we have access to the model tokenizer via the LLM object defined above lets use that here?

src/autolabel/models/vllm.py
```diff
@@ -0,0 +1,93 @@
...
+                        error_type=ErrorType.LLM_PROVIDER_ERROR, error_message=str(e)
+                    )
+                )
+                latencies.append(0)
+        return RefuelLLMResult(generations=generations, errors=errors, latencies = latencies)
+
+    def get_cost(self, prompt: str, label: Optional[str] = "") -> float:
+        return 0
+
+    def returns_token_probs(self) -> bool:
```
DhruvaBansal00, line 82: I think this can return logprobs, right?

benchmark/benchmark.py
rajasbansal: fixed

src/autolabel/models/vllm.py
rajasbansal: not adding this to requirements as this is experimental. but still fixed how this would be used

DhruvaBansal00 (approved):

src/autolabel/tasks/question_answering.py
```diff
@@ -31,9 +31,7 @@
 
 
 class QuestionAnsweringTask(BaseTask):
-    DEFAULT_OUTPUT_GUIDELINES = (
-        'You will return the answer one element: "the correct label"\n'
-    )
```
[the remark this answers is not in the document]
rajasbansal: Yep! That prompt changed the performance of gpt-4-turbo a lot

PR #779, opened 2024-04-01
Use llama prompt schema when model_name is set

rajasbansal:
# Pull Review Summary

## Description

Mistral finetuning

## Type of change

## Tests

locally

### DIFF
--- a/src/autolabel/tasks/base.py
+++ b/src/autolabel/tasks/base.py
@@ -69,17 +69,10 @@ class BaseTask(ABC):
         self._prompt_schema_init()
 
     def _prompt_schema_init(self) -> None:
-<<<<<<< Updated upstream
-        self.use_llama_prompt_schema = self.config.provider() in [
-            ModelProvider.REFUEL,
-            ModelProvider.TGI,
-        ]
-=======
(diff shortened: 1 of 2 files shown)


src/autolabel/tasks/base.py
```diff
@@ -67,10 +69,10 @@ def __init__(self, config: AutolabelConfig) -> None:
...
 
     def _prompt_schema_init(self) -> None:
-        self.use_llama_prompt_schema = self.config.provider() in [
-            ModelProvider.REFUEL,
-            ModelProvider.TGI,
-        ]
+        self.use_llama_prompt_schema = (
+            self.config.provider() == ModelProvider.REFUEL
+            and self.config.model_name() == REFUEL_LLM_MODEL
+        ) or self.config.provider() == ModelProvider.TGI
```
DhruvaBansal00, line 75: We currently concatenate options in the classification task differently based on the model provider as well. We might need a similar change there too.

DhruvaBansal00 (approved):

src/autolabel/tasks/base.py
rajasbansal: thanks fixing this

PR #780, opened 2024-04-01
Add premium proxy on webpage transform retries

Abhinav-Naikawadi:
# Pull Review Summary

## Description

We currently run into errors when scraping certain domains. When we get an error on such requests, we will retry with a proxy to reduce the chance of error.

## Type of change

## Tests

Tested retry logic with proxy

### DIFF
--- a/src/autolabel/transforms/webpage_transform.py
+++ b/src/autolabel/transforms/webpage_transform.py
@@ -10,6 +10,7 @@ import asyncio
 import logging
 import pandas as pd
 import ssl
+import json
 from langchain_community.document_transformers import Html2TextTransformer
 from langchain.docstore.document import Document
 from autolabel.cache import BaseCache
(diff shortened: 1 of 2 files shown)


src/autolabel/transforms/webpage_transform.py
```diff
@@ -55,9 +54,13 @@ async def _load_url(self, url: str, retry_count=0) -> str:
             raise TransformError(
                 TransformErrorType.MAX_RETRIES_REACHED, "Max retries reached"
             )
+        if retry_count > 0:
```
yadavsahil197, line 57: if we know specific codes to retry using premium proxy, lets do that without doing a backoff

src/autolabel/transforms/webpage_transform.py
```diff
@@ -55,9 +54,13 @@ async def _load_url(self, url: str, retry_count=0) -> str:
...
+            logger.warning(f"Retrying scraping URL: {url} with premium proxy")
```
yadavsahil197, line 58: nit: info

yadavsahil197 (approved):
LG otherwise

PR #781, opened 2024-04-01
Misleading log for non-classification tasks

Abhinav-Naikawadi:
# Pull Review Summary

## Description

The log is misleading, since labels list is only relevant for classification tasks.

## Type of change

## Tests

Removing log


Vaibhav2001 (approved):
Yeah I had that for experimentation, I mistakenly pushed this I guess. Removing this should be fine

PR #794, opened 2024-04-09
Fix pricing for gemini models

Vaibhav2001:
# Pull Review Summary

## Description

Fix pricing for gemini models based on number of tokens.

## Type of change

### DIFF
--- a/src/autolabel/models/google.py
+++ b/src/autolabel/models/google.py
@@ -112,9 +112,9 @@ class GoogleLLM(BaseModel):
             return 0.0
         cost_per_prompt_token = self.COST_PER_PROMPT_TOKEN[self.model_name]
         cost_per_completion_token = self.COST_PER_COMPLETION_TOKEN[self.model_name]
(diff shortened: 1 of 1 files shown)


src/autolabel/models/google.py
```diff
@@ -73,6 +74,7 @@ def __init__(
         model_params = config.model_params()
         self.model_params = {**self.DEFAULT_PARAMS, **model_params}
         self.llm = VertexAI(model_name=self.model_name, **self.model_params)
+        self.vertexaiModel = generative_models.GenerativeModel(self.model_name)
```
rajasbansal, line 77: hope this doesnt make some connection to the llm api as well

rajasbansal (approved):

src/autolabel/models/google.py
Vaibhav2001: I do not know if it makes a connection but it does make a post request but according to @nihit it was fine. Maybe I am missing something?

rajasbansal: sg

PR #795, opened 2024-04-11
Confidence score fallback

Abhinav-Naikawadi:
# Pull Review Summary

## Description

When OpenAI does not return logprobs, we fallback to using RefuelLLM for confidence scores. This would mean that the confidence scores are not normalized with OpenAI logprob confidence scores, so this might not be ideal, but perhaps is better than returning 0 confidence

## Type of change

## Tests

Tested locally for case when no logprobs are present (manually set logprobs to None)


rajasbansal:
@Abhinav-Naikawadi lets close this

PR #796, opened 2024-04-11
Fix transform cache entry

Abhinav-Naikawadi:
# Pull Review Summary

## Description

Previously, the transform cache entry was relying on the input row dict as one of the fields we include when computing the hash. However, this is incorrect behavior because we should only be including the items in the row corresponding to the input columns of the transform (i.e. a url column for a webpage scrape or the query columns for a web search transform). The old behavior was leading to errors when the dataframe used in labeling contained columns that had non-serializable values (for instance, when there was an LLMAnnotation column). Furthermore, in cases where the remainder of the row changes (very possible with the introduction of task chaining, where the transform might be a subtask downstream of previous labeling), but the actual value corresponding to the transform input does not change (i.e. the URL column), then we would previously have been getting cache misses, even though the transform was already computed and we can still use the cache result.

## Type of change

## Tests

Tested locally with transform cache


yadavsahil197 (approved):
Do we need the params property anymore? If not, let's remove it.

PR #800, opened 2024-04-15
Added support for Mistral

Vaibhav2001:
# Pull Review Summary

## Description

Added support for Mistral

## Type of change

## Tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,7 +88,6 @@ minimal = [
     "langchain-community==0.0.27",
     "langchain-openai==0.0.7",
     "langchain_google_vertexai==0.0.5",
-    "langchain-google-genai==0.0.9",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -88,7 +88,6 @@ minimal = [
     "langchain-community==0.0.27",
     "langchain-openai==0.0.7",
     "langchain_google_vertexai==0.0.5",
-    "langchain-google-genai==0.0.9",
```
rajasbansal, line 91: do we not need this now?

@yadavsahil197 does removal need some extra work now since we are generating the lockfile?

rajasbansal (approved):

pyproject.toml
Vaibhav2001: No since we only use vertex-ai

PR #822, opened 2024-05-10
Optionally return LLM annotations for explanations

Abhinav-Naikawadi:
# Pull Review Summary

## Description

Return LLM annotations with metrics for explanation generation optionally.

## Type of change

## Tests

Tested locally

### DIFF
--- a/src/autolabel/labeler.py
+++ b/src/autolabel/labeler.py
@@ -637,7 +637,7 @@ class LabelingAgent:
                     annotation = LLMAnnotation(
                         successfully_labeled=False,
                         label=self.task.NULL_LABEL_TOKEN,
-                        raw_response="",
+                        raw_response=explanation,
                         curr_sample=pickle.dumps(seed_example),
                         prompt=explanation_prompt,
                         confidence_score=0,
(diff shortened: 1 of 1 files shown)


src/autolabel/labeler.py
```diff
@@ -584,18 +584,22 @@ def generate_explanations(
         self,
         seed_examples: Union[str, List[Dict]],
         include_label: bool = True,
+        return_annotations: bool = False,
```
rajasbansal, line 587: should we always just return annotations? might make this a bit cleaner

src/autolabel/labeler.py
```diff
@@ -623,14 +627,45 @@ async def agenerate_explanations(
...
+            explanation = response.generations[0][0].text
             seed_example[explanation_column] = str(explanation) if explanation else ""
+            if return_annotations:
+                error = response.errors[0]
+                input_tokens = self.llm.get_num_tokens(explanation_prompt)
+                if error is not None:
+                    annotation = LLMAnnotation(
+                        successfully_labeled=False,
+                        label=self.task.NULL_LABEL_TOKEN,
+                        raw_response="",
```
rajasbansal, line 640: we should still return the explanation from above potentially?

rajasbansal (approved):

PR #832, opened 2024-05-24
Refuel LLM: Use endpoint from config

yadavsahil197:
# Pull Review Summary

## Description

Passing the endpoint in the model config.

## Type of change

## Tests

Tested locally

### DIFF
--- a/src/autolabel/configs/config.py
+++ b/src/autolabel/configs/config.py
@@ -34,6 +34,7 @@ class AutolabelConfig(BaseConfig):
     PROVIDER_KEY = "provider"
     MODEL_NAME_KEY = "name"
     MODEL_PARAMS_KEY = "params"
+    MODEL_ENDPOINT_KEY = "endpoint"
     COMPUTE_CONFIDENCE_KEY = "compute_confidence"
     LOGIT_BIAS_KEY = "logit_bias"
     JSON_MODE = "json_mode"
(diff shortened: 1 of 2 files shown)


src/autolabel/configs/config.py
```diff
@@ -169,6 +170,10 @@ def model_params(self) -> Dict:
         """Returns a dict of configured settings for the model (e.g. hyperparameters)"""
         return self._model_config.get(self.MODEL_PARAMS_KEY, {})
 
+    def model_endpoint(self) -> str:
+        """Returns the endpoint to use for the model"""
+        return self._model_config.get(self.MODEL_ENDPOINT_KEY, None)
```
DhruvaBansal00, line 175: @yadavsahil197 to ensure this code change works for RefuelLLMv2 large and small models, we'll have to change the elb endpoint (to reflect llm.refuel.ai/models/.../v2/generate) in their LLMEntity tables, correct?

yadavsahil197: The base model configs already have the updated endpoints.
For finetuned models, we will need to set the elb endpoint to the correct values.

yadavsahil197: Also, this doesn't change the behavior for realtime requests, as the inference server continues to use the earlier deployed model.

DhruvaBansal00 (approved):

PR #833, opened 2024-05-24
Search for valid JSON in response text on LLM Parsing error

Abhinav-Naikawadi:
# Pull Review Summary

## Description

When an LLM parsing error is encountered for attribute extraction tasks, attempt to find the valid JSON in the response text.

## Type of change

## Tests

Tested locally for responses of the form: Based on the information, the output is {"": ""}

### DIFF
--- a/src/autolabel/tasks/attribute_extraction.py
+++ b/src/autolabel/tasks/attribute_extraction.py
@@ -215,11 +215,24 @@ class AttributeExtractionTask(BaseTask):
             llm_label = {k: str(v) for k, v in json.loads(completion_text).items()}
             successfully_labeled = True
         except Exception as e:
-            logger.error(f"Error parsing LLM response: {response.text}, Error: {e}")
-            llm_label = self.NULL_LABEL
-            error = LabelingError(
-                error_type=ErrorType.PARSING_ERROR, error_message=str(e)
+            logger.info(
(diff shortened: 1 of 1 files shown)


src/autolabel/tasks/attribute_extraction.py
```diff
@@ -215,11 +215,24 @@ def parse_llm_response(
...
-            logger.error(f"Error parsing LLM response: {response.text}, Error: {e}")
-            llm_label = self.NULL_LABEL
-            error = LabelingError(
-                error_type=ErrorType.PARSING_ERROR, error_message=str(e)
+            logger.info(
+                f"Error parsing LLM response: {response.text}, Error: {e}. Now searching for valid JSON in response"
             )
+            try:
+                json_start, json_end = response.text.find("{"), response.text.rfind("}")
+                llm_label = {
```
rajasbansal, line 223: how about we remove the earlier way of loading the json and just find the json string in the output always? the check before this which removes markdown seems to be a subset of finding the brackets anyway

rajasbansal (approved):

src/autolabel/tasks/attribute_extraction.py
Abhinav-Naikawadi: for now, leaving as is to log when the output is ill-formatted (even if the valid JSON substring search works for some examples, might be good to adjust prompt if this is necessary to begin it)

PR #843, opened 2024-06-04
custom api transform support

Abhinav-Naikawadi:
# Pull Review Summary

## Description

New custom API transform support

## Type of change

## Tests

Tested locally

### DIFF
--- a/src/autolabel/transforms/custom_api.py
+++ b/src/autolabel/transforms/custom_api.py
@@ -76,7 +76,7 @@ class CustomApi(BaseTransform):
         return TransformType.CUSTOM_API
 
     async def _get_result(
-        self, url: str, verify=True, headers=HEADERS, retry_count=0
+        self, url: str, params: Dict, verify=True, headers=HEADERS, retry_count=0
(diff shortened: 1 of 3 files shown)


src/autolabel/transforms/custom_api.py
```diff
@@ -0,0 +1,136 @@
...
+    async def _apply(self, row: Dict[str, Any]) -> Dict[str, Any]:
+        for col in self.request_columns:
+            if col not in row:
+                logger.error(
+                    f"Missing request column: {col} in row {row}",
+                )
+        request_template = f"{self.base_url}?" + "&".join(
+            [f"{col}={{{col}}}" for col in self.request_columns if col in row]
+        )
+        url = request_template.format_map(defaultdict(str, row))
```
nihit, line 122: let's use a standardized library like requests or similar to do this - we need to make sure parameter values are urlencoded etc.
```
r = requests.get(base_url, params={'key': 'value', ...})
print(r.url)
```

nihit (approved):
lgtm otherwise

PR #849, opened 2024-06-07
Return better error messages and types for labeling errors

Abhinav-Naikawadi:
# Pull Review Summary

## Description

Return better error messages and types for labeling errors by parsing LLM provider errors for appropriate error types when available. Still use default error types as fallback.

## Type of change

## Tests

Tested locally

### DIFF
--- a/src/autolabel/models/openai.py
+++ b/src/autolabel/models/openai.py
@@ -1,4 +1,5 @@
 import copy
+import ast
 import logging
 import os
 from functools import cached_property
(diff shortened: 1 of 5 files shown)


src/autolabel/models/openai.py
```diff
@@ -221,12 +226,27 @@ async def _alabel(self, prompts: List[str]) -> RefuelLLMResult:
             )
         except Exception as e:
             logger.exception(f"Unable to generate prediction: {e}")
+            error_message = str(e)
```
Abhinav-Naikawadi, line 229: Added new error parsing logic for OpenAI here. If we want to land these changes, can also add similar logic for other providers.

nihit (approved):
lgtm otherwise

PR #854, opened 2024-06-17
Serp api query if input columns are empty

rajasbansal:
# Pull Review Summary

## Description

Serp api query if input columns are empty. Currently, if the inputs are empty i.e None, we send the string None to google. We should probably change that to an empty string by explicitly filtering this dictionary

## Type of change

### DIFF
--- a/src/autolabel/transforms/serp_api.py
+++ b/src/autolabel/transforms/serp_api.py
@@ -105,7 +105,9 @@ class SerpApi(BaseTransform):
                 logger.error(
                     f"Missing query column: {col} in row {row}",
                 )
-        query = self.query_template.format_map(defaultdict(str, row))
+        query = self.query_template.format_map(
+            defaultdict(str, {key: val for key, val in row.items() if val is not None})
+        )
         search_result = self.NULL_TRANSFORM_TOKEN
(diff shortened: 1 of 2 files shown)


src/autolabel/transforms/serp_api.py
```diff
@@ -105,7 +105,9 @@ async def _apply(self, row: Dict[str, Any]) -> Dict[str, Any]:
                 logger.error(
                     f"Missing query column: {col} in row {row}",
                 )
-        query = self.query_template.format_map(defaultdict(str, row))
+        query = self.query_template.format_map(
```
yadavsahil197, line 108: can we also handle the case when all input columns are empty?

rajasbansal: what does that mean? The default dict changes all empty columns to ""

Abhinav-Naikawadi (approved):
makes sense, lgtm after addressing other comment!

src/autolabel/transforms/serp_api.py
Abhinav-Naikawadi: I think perhaps we don't make query if no columns present in row?

PR #861, opened 2024-06-26
Validate attribute extraction labels list

Abhinav-Naikawadi:
# Pull Review Summary

## Description

Validate attribute extraction labels list

## Type of change

## Tests

Tested locally

### DIFF
--- a/src/autolabel/tasks/attribute_extraction.py
+++ b/src/autolabel/tasks/attribute_extraction.py
@@ -238,18 +238,13 @@ class AttributeExtractionTask(BaseTask):
         if successfully_labeled:
             for attribute in self.config.attributes():
                 attr_options = attribute.get("options")
-                if attr_options is not None:
+                if attr_options is not None and len(attr_options) > 0:
                     attr_label = llm_label.get(attribute["name"])
                     if attr_label is not None and attr_label not in attr_options:
                         logger.warning(
                             f"Attribute {attr_label} from the LLM response {llm_label} is not in the labels list"
                         )
-                        successfully_labeled = False
-                        error = LabelingError(
-                            error_type=ErrorType.OUTPUT_GUIDELINES_NOT_FOLLOWED_ERROR,
-                            error_message=f"LLM response is not in the labels list.",
-                        )
-                        llm_label = self.NULL_LABEL_TOKEN
+                        llm_label.pop(attribute["name"], None)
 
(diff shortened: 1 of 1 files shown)


src/autolabel/tasks/attribute_extraction.py
```diff
@@ -235,7 +235,21 @@ def parse_llm_response(
...
+                if attr_options is not None:
+                    attr_label = llm_label.get(attribute["name"])
+                    if attr_label is not None and attr_label not in attr_options:
+                        logger.warning(
+                            f"Attribute {attr_label} from the LLM response {llm_label} is not in the labels list"
+                        )
+                        successfully_labeled = False
+                        error = LabelingError(
+                            error_type=ErrorType.OUTPUT_GUIDELINES_NOT_FOLLOWED_ERROR,
+                            error_message=f"LLM response is not in the labels list.",
```
rajasbansal, line 250: also output what the raw response was in the error message?

src/autolabel/tasks/attribute_extraction.py
```diff
@@ -235,7 +235,21 @@ def parse_llm_response(
...
+                        )
+                        llm_label = self.NULL_LABEL_TOKEN
```
rajasbansal, line 252: should we just change this particular subtask to None instead of changing the entire llm_label

src/autolabel/tasks/attribute_extraction.py
```diff
@@ -235,7 +235,16 @@ def parse_llm_response(
...
                 )
 
-        # TODO(rajas): Handle output guidelines not followed error (for options case)
+        if successfully_labeled:
+            for attribute in self.config.attributes():
+                attr_options = attribute.get("options")
+                if attr_options is not None and len(attr_options) > 0:
+                    attr_label = llm_label.get(attribute["name"])
+                    if attr_label is not None and attr_label not in attr_options:
+                        logger.warning(
```
rajasbansal, line 244: not returning an error?

rajasbansal (approved):

src/autolabel/tasks/attribute_extraction.py
Abhinav-Naikawadi: yeah since errors are at an LLMAnnotation level, it would require either returning no label, or filtering out the invalid attributes (as is done here).

PR #887, opened 2024-08-13
Validate example output dict keys in few shot examples

Abhinav-Naikawadi:
# Pull Review Summary

## Description

This change will validate example output dict keys in few shot examples. Any few shot example that does not have labels for all expected output fields will be filtered out during prompt construction.

## Type of change

## Tests

Tested locally with tasks that did and did not have appropriate few shot examples with labels

### DIFF
--- a/src/autolabel/tasks/attribute_extraction.py
+++ b/src/autolabel/tasks/attribute_extraction.py
@@ -90,7 +90,28 @@ class AttributeExtractionTask(BaseTask):
         for attribute in self.config.attributes():
             attribute_name = attribute["name"]
             output_dict[attribute_name] = input.get(attribute_name, "")
+        if not self._validate_output_dict(output_dict):
+            logger.warn(
+                f"Generated output dict: {output_dict} does not contain all the expected output attributes. Skipping example."
+            )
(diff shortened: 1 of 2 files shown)


src/autolabel/tasks/attribute_extraction.py
```diff
@@ -118,6 +139,8 @@ def construct_prompt(
         for eg in examples:
```
DhruvaBansal00, line 139: Just confirming - this effectively means that if we had 5 examples passed in and 1 one of them doesn't have all GT labels populated, we'll now only send 4 examples in? We don't explicitly go and fetch a new example to ensure that 5 are going in, correct? 
I think this behavior is fine as is, wanted to make sure I am understanding this correctly.

DhruvaBansal00 (approved):

src/autolabel/tasks/attribute_extraction.py
Abhinav-Naikawadi: Yeah, we do plan to revamp example selection/vector search to improve on this behavior, but that will come as a followup.

PR #892, opened 2024-08-24
Add support for num results web search parameter

Abhinav-Naikawadi:
# Pull Review Summary

## Description

Add support for num results web search parameter

## Type of change

- 
## Tests

Tested locally with tasks with a web search enrichment


src/autolabel/transforms/serper_api.py
```diff
@@ -67,6 +64,7 @@ def __init__(
         query_template: str,
         serper_api_key: str,
         serper_args: dict = DEFAULT_ARGS,
+        num_results: int = 10,
```
nihit, line 67: let's clean up this interface - we can unify all of these under kwargs and pass it to GoogleSerperAPIWrapper, or have a separate argument for each. Instead of the current implementation where some parameters are in `serper_args` and `num_results` is its own parameter

nihit (approved):
lgtm otherwise

PR #894, opened 2024-08-29
Validate webpage scrape URLs

Abhinav-Naikawadi:
# Pull Review Summary

## Description

Validate webpage scrape URLs using regex. The motivation is to avoid making web scrape requests unnecessarily when the URL is invalid. It is possible that there may be some URL's for which this regex match does not validate correctly. However, these cases should be very niche and this change will avoid several unnecessary web scrape requests for a number of cases where the URL is obviously invalid.

## Type of change

## Tests

Tested locally with common valid and invalid URL's/patterns

### DIFF
--- a/src/autolabel/transforms/webpage_scrape.py
+++ b/src/autolabel/transforms/webpage_scrape.py
@@ -146,7 +146,7 @@ class WebpageScrape(BaseTransform):
             if not validate_url(url):
                 raise TransformError(
                     TransformErrorType.INVALID_INPUT,
-                    f"Invalid url in row {row}",
+                    f"Invalid url: {url}",
                 )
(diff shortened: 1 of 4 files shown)


src/autolabel/transforms/webpage_scrape.py
```diff
@@ -142,6 +143,11 @@ async def _apply(self, row: Dict[str, Any]) -> Dict[str, Any]:
         else:
             if not urlparse(url).scheme:
                 url = f"https://{url}"
+            if not validate_url(url):
```
Abhinav-Naikawadi, line 146: it might be unnecessary to do this for this webscrape transform

nihit (approved):

PR #908, opened 2024-09-30
Allow label selection on attribute extraction tasks

rajasbansal:
Allows using label selection if specified for attribute extraction tasks


src/autolabel/few_shot/vector_store.py
```diff
@@ -17,13 +17,6 @@
 
 logger = logging.getLogger(__name__)
 
-try:
-    import torch
```
nihit, line 21: should we remove this requirement from pyproject.toml ?

nihit (approved):

PR #918, opened 2024-10-09
GPT 4o now supports structured outputs

DhruvaBansal00:


tuxracer (approved):

nihit (approved):
on a related note - we need update the per token pricing for gpt-4o

PR #927, opened 2024-10-25
Pass tokenizer from LabelingAgent to llm

yadavsahil197:
Reusing the tokenizer reduces the latency.

### DIFF
--- a/src/autolabel/models/google.py
+++ b/src/autolabel/models/google.py
@@ -1,9 +1,10 @@
-import os
 import logging
+import os
 from time import time
 from typing import Dict, List, Optional
 
 from langchain.schema import Generation
+from transformers import AutoTokenizer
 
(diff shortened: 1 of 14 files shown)


nihit (approved):

src/autolabel/models/google.py
```diff
@@ -34,13 +35,14 @@ def __init__(
         self,
         config: AutolabelConfig,
         cache: BaseCache = None,
+        tokenizer: Optional[AutoTokenizer] = None,
```
rajasbansal, line 38: these models dont use the tokenizer at all

rajasbansal (approved):

src/autolabel/models/google.py
yadavsahil197: No, they don't. But we need as its part of the interface

PR #935, opened 2024-11-14
Use aws textract for ocr

DhruvaBansal00:


src/autolabel/transforms/ocr.py
```diff
@@ -1,110 +1,199 @@
...
         try:
-            from PIL import Image
             import pytesseract
             from pdf2image import convert_from_path
 
-            self.Image = Image
             self.pytesseract = pytesseract
             self.convert_from_path = convert_from_path
             self.pytesseract.get_tesseract_version()
+
```
rajasbansal, line 60: probably dont need pytesseract for ocr now?

rajasbansal (approved):

src/autolabel/transforms/ocr.py
DhruvaBansal00: Removed

PR #945, opened 2024-12-11
Allow sending label descriptions to the LLM

rajasbansal:


src/autolabel/tasks/attribute_extraction.py
```diff
@@ -107,6 +108,20 @@ def _construct_attribute_json(
...
+                if (
+                    selected_labels_desc_map
+                    and attribute_name in selected_labels_desc_map
+                ):
+                    attribute_options_desc = selected_labels_desc_map[attribute_name]
+                attribute_options_desc = {
+                    k: v for k, v in attribute_options_desc.items() if v is not None
+                }
+                if attribute_options_desc:
+                    attribute_desc += "\nDescription for each option:"
```
nihit, line 121: do we need to modify anything in the prompt to dissuade the LLM to output anything other than the option? i.e. so it does not output anything from the description

nihit: if we have tested this for some usecases already, that's great

nihit (approved):

src/autolabel/tasks/attribute_extraction.py
rajasbansal: Yep tried it on a couple of usecases, separating out the description and the options seems to make sure that the LLM does not output the description as well.

